PR #69, opened 2021-06-26
Improve Emoji Performance

### LINKED ISSUE #17: Major lag spikes during Emoji spam
Should create a better Emoji implementation, current one is very inefficient

AnatoleAM:
Creating various improvements to the performance of emoji rendering, by removing unnecessary/duplicate calls to the json index file.

This aims to fix #17


AnatoleAM:
While still an imperfect implementation, by no longer turning emojis into emote components and removing calls to the massive JSON emoji data file, performance is boosted significantly to a point where it should no longer cause visible lag to the end user.

PR #90, opened 2021-08-06
Inline Chat Preview

AnatoleAM:
The creation of a new feature adding the ability to view emotes as they are typed in the chat bar


AnatoleAM:

PR #98, opened 2021-08-09
Better tab-complete

Melonify:
* Allow the ability to tab complete words other than the last one
* Reset tab complete cursor when input has been changed
* Finish tab complete when the user unfocuses the chat box


AnatoleAM:
cum

AnatoleAM (approved):
According to all known laws
of aviation,

there is no way a bee
should be able to fly.

Its wings are too small to get
its fat little body off the ground.

The bee, of course, flies anyway
  
because bees don't care
what humans think is impossible.

PR #139, opened 2021-09-19
Menu button fixes

Excellify:
Fixed an issue with the emote menu being a few pixels too large and not clickable on the sides outside the logo.
Made the 7TV settings button have a background on hover.

### DIFF
--- a/src/Sites/app/EmbeddedUI.tsx
+++ b/src/Sites/app/EmbeddedUI.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { EmoteMenuButton } from 'src/Sites/app/EmoteMenu/EmoteMenuButton';
-import { assetStore, SiteApp } from 'src/Sites/app/SiteApp';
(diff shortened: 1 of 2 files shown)


AnatoleAM (commented):
This is good, but is there a reason the class name was changed?

Excellify:
Changed the name because i used the style for both the emote menu button and the settings button

AnatoleAM:
Oh right that's correct, this is used in both the navigation bar and chat so the change makes sense.

AnatoleAM (approved):

Excellify:
Also looked at adding a tooltip to the settings button, similarly to the other buttons up there. And maybe making the open settings button toggle the settings window.

AnatoleAM:
I think a tooltip would require making a custom one, I haven't figured out how to access twitch's tooltip

Excellify:
A very similar tooltip to twitch's can be made with just some css, started messing around with it but didn't get it done. FFZ's tooltip is not identical to twitch's either.

PR #140, opened 2021-09-20
Adding tooltip to menu buttons 

Excellify:
Im sure there are prettier ways of making a tooltip in css, but this works at least.

* Added tooltips to menu buttons

* Removed whitespace


Excellify:
Still learning git, so my commit's are all over the place...

Excellify:
Just realized lots of this messes with how it looks on youtube. Will fix tomorrow...

PR #148, opened 2021-09-25
Double @ fix

Excellify:
Fixes the Double @ when having both 7tv and ffz installed. The historical view of messages and the first message would have a double @'User'. This would only happen in channels where you are a moderator. The reason was that the ffzMode variable was set at when the currentHandler.start() was first called. At this point the ffz hook hasn't loaded yet, so it would be set to false. Later when the first message is sendt and the componentDidUpdate() is called, the local ffzMode is still set to false, at which point it should be true.


Excellify:
Apparently being a moderator is not required to for the issue to occur.

AnatoleAM:
Is this actually working? @Excellify

Excellify:
it fixed the double @ when joining a channel and loading in messages from history at least

PR #168, opened 2021-10-19
True theatre mode improvements

Excellify:
Fixes offset when viewing vertical youtube streams in true theatre mode.
Improves the divisor a bit.


Excellify:
The bug you experienced was is there in the current version i believe, the latest commit fixes it i believe.

AnatoleAM (approved):

PR #172, opened 2021-10-29
Fix youtube input box not updating after adding emote from menu

Excellify:
This fixes youtube's chat input box not updating after adding an emote.
Users no longer have to add a space after adding an emote for the chat input box to update.


AnatoleAM (approved):
👍

PR #173, opened 2021-10-29
Fix Clickable hyperlinks in youtube chat

Excellify:
Renders links as clickable when the sender is a moderator in the channel.

### DIFF
--- a/src/Sites/youtube.com/Runtime/Tokenizer.ts
+++ b/src/Sites/youtube.com/Runtime/Tokenizer.ts
@@ -80,6 +80,7 @@ export class Tokenizer {
 		a.innerText = text;
 		a.className = 'yt-simple-endpoint style-scope yt-live-chat-text-message-renderer';
 		a.href = text;
+		a.target = '_blank';
 
 		if (a.host && a.host != window.location.host) {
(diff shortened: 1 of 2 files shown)


AnatoleAM (changes requested):
One small problem

src/Sites/youtube.com/Runtime/Tokenizer.ts
```diff
@@ -74,6 +72,22 @@ export class Tokenizer {
 		ctx.appendChild(span);
 	}
 
+	/**
+	 * Append a text part to the new message context
+	 */
+	addHyperlinkPart(ctx: HTMLSpanElement, text: string): boolean {
```
AnatoleAM, line 78: This doesn't  open in a new tab, so if a user clicks a link they redirect to the page and quit the stream

AnatoleAM (changes requested):

src/Sites/youtube.com/Runtime/Tokenizer.ts
Excellify: Oops, ill fix that

AnatoleAM (approved):
👍

PR #175, opened 2021-10-30
Fix Unload slider on message unmount

Excellify:
Unloads the ban slider when the message is unloaded. 
This is to avoid memory leaks.


AnatoleAM (approved):
Seems good!

PR #177, opened 2021-11-01
Fix youtube popout wrong channel id

Excellify:
Fixed an issue where the current users channel id was used for channel emotes instead of the channel's.

### DIFF
--- a/src/Sites/youtube.com/youtube.tsx
+++ b/src/Sites/youtube.com/youtube.tsx
@@ -176,7 +176,6 @@ export class YouTubePageScript {
 				.sendLiveChatMessageEndpoint
 				.params;
 
-			//Copied from bttv
 			channelID = atob(decodeURIComponent(atob(channelID)))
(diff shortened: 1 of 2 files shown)


src/Sites/youtube.com/youtube.tsx
```diff
@@ -163,22 +163,26 @@ export class YouTubePageScript {
...
-				.actions[0]
-				.addLiveChatTextMessageFromTemplateAction
-				.template
-				.liveChatTextMessageRenderer;
+				.params;
+
+			channelID = atob(decodeURIComponent(atob(channelID)))
+			.split(`*'\n\u0018`)[1]
+			.split('\u0012\u000b')[0];
+
```
AnatoleAM, line 182: Is this removing the regular fallback? I believe this will break implementation in certain views, such as watching a VOD

AnatoleAM (approved):

PR #180, opened 2021-11-03
Fix Bad Imports

AnatoleAM:
Fixing bad imports of `material-ui`, which were importing the entire package in several instances, causing the output bundle size to swell into ridiculous sizes.

@Excellify 🤨


src/Sites/twitch.tv/Components/BanSlider.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { PointerEvent, useState } from 'react';
-import { Restore } from '@material-ui/icons';
-import { SvgIcon } from '@material-ui/core';
+import Restore from '@material-ui/icons/Restore';
+import SvgIcon from '@material-ui/core/SvgIcon';
```
Excellify, line 4: Only this one was my doing right?

PR #191, opened 2021-12-24
Fix paints on firefox

Excellify:
Fixes paints on firefox. 
Firefox dosnt allow you to access stylesheets that come from a different source than the origin. This includes all extensions as well as our own stylesheet for some reason.
Creates and returns a new stylesheet if the user is using firefox.


AnatoleAM (approved):
Thank you!

PR #195, opened 2022-01-01
Fix paints style builder

Excellify:
Fixes a missing semicolon in the paints style builder that caused the color and drop shadow to not work when a color was included in the paint.


AnatoleAM (approved):
Thank you

PR #196, opened 2022-01-03
Added more modifiers

LeptailurusServal:
Effects:
-  Poof
- Arrive
- Leave
- SpinY
- SpinTOODANK


Excellify:
Merging this to remove the pull request, but keep in mind that it's unlikely the modifier feature will make it to the extension at all. At least in its current form.

PR #198, opened 2022-01-08
Added support for non-ascii names

Excellify:
Added support for non ascii names by moving the paints css class one element back and removing the opacity on the ascii name. 
Also made static parts of the paints style its own css style


AnatoleAM:
Is this complete? @Excellify

Excellify:
> 

Should be good

AnatoleAM (approved):

PR #199, opened 2022-01-12
Added support for auto completing emotes with :

Excellify:
This adds all emotes to twitch's : auto-complete menu. Should be compatible with ffz, but haven't been able to test extensively.


AnatoleAM (approved):
Banger

PR #201, opened 2022-01-16
Colon-complete fixes

Excellify:
Fixes 2 issues with the colon-complete: 

1. Will no longer throw an error when recieving "cease" from ffz if the TwitchPageScript hasn't been initialized yet.
2. Colon-complete will now wait if twitch has not finished inserting its own emotes.

### DIFF
--- a/src/Sites/twitch.tv/twitch.tsx
+++ b/src/Sites/twitch.tv/twitch.tsx
@@ -145,8 +145,14 @@ export class TwitchPageScript {
 	insertEmotesIntoAutocomplete() {
 		if (this.ffzMode) return;
 
-		const emoteProvider = this.twitch.getAutocompleteHandler().providers[0];
 		const store = this.emoteStore;
+		const emoteProvider = this.twitch.getAutocompleteHandler().providers[0];
+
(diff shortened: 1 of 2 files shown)


src/Sites/twitch.tv/twitch.tsx
```diff
@@ -204,7 +206,7 @@ export class TwitchPageScript {
 	whenUpperLayerRequestsThePageScriptStopsSendingChatLinesUpstream(): void {
 		ffzMode = true;
 
-		let sets = this.twitch.getAutocompleteHandler()?.providers[0].props.emotes;
+		let sets = this.twitch?.getAutocompleteHandler()?.providers[0].props.emotes;
```
AnatoleAM, line 209: `this.twitch` will always be undefined, since the decorator is calling this method which doesn't pass the context properly. This doesn't seem like the correct place to put this

AnatoleAM (changes requested):

AnatoleAM (approved):
I applied a change moving the `Twitch` instance out of class scope. Unfortunately, the method decorator approach we have doesn't seem to be very functional

PR #210, opened 2022-01-20
click to use feature

ddenniem:
This feature adds functionality to click any emote in chat and it will input itself into the chat draft, as if it was an emote menu injection.

### DIFF
--- a/src/Content/Runtime/Settings.ts
+++ b/src/Content/Runtime/Settings.ts
@@ -117,6 +117,13 @@ class Settings {
 					defaultValue: true
 				},
 				{
+					id: 'general.emote_click',
+					label: 'Click emote to use',
+					hint: 'Click on any emote in chat to automatically input it in your chat draft',
+					type: 'checkbox',
+					defaultValue: true
+				},
+				{
 					id: 'general.history_navigation',
 					label: 'History Navigation',
 					hint: 'Allow the usage of up/down arrows to get previously sent messages back in the chat input box',
--- a/src/Content/Runtime/Settings.ts
+++ b/src/Content/Runtime/Settings.ts
@@ -106,7 +106,7 @@ class Settings {
 					label: 'Click emote to use',
 					hint: 'Click on any emote in chat to automatically input it in your chat draft',
 					type: 'checkbox',
-					defaultValue: true
+					defaultValue: false
 				},
 				{
 					id: 'general.app_avatars',
--- a/src/Content/Runtime/Settings.ts
+++ b/src/Content/Runtime/Settings.ts
@@ -103,8 +103,8 @@ class Settings {
 				},
 				{
 					id: 'general.emote_click',
-					label: 'Click emote to draft',
-					hint: 'Click on any emote in chat to automatically input it in your chat draft',
+					label: 'Shift-Click emote to draft',
+					hint: 'Shift + Click on any emote in chat to automatically input it into your chat draft',
 					type: 'checkbox',
 					defaultValue: false
 				},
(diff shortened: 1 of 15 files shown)


ddenniem:
I implemented a setting for the added feature, but in my testing, the setting does not disable the feature when turned off. I tested on firefox.

src/Global/EmoteStore.tsx
```diff
@@ -400,6 +402,12 @@ export namespace EmoteStore {
 				MainComponent.ShowTooltip.next({ event, hover: false });
 			});
 
+			inner.addEventListener('mousedown', _ => {
+				if(!settings.get('general.emote_click').asBoolean()) return;
```
Excellify, line 406: Use the configMap from SiteApp instead of settings for this to work.

src/Content/Runtime/Settings.ts
```diff
@@ -101,6 +101,13 @@ class Settings {
 					type: 'checkbox',
 					defaultValue: true
 				},
+				{
+					id: 'general.emote_click',
+					label: 'Click emote to use',
+					hint: 'Click on any emote in chat to automatically input it in your chat draft',
+					type: 'checkbox',
+					defaultValue: true
```
Excellify, line 109: Unsure if this should be enabled by default as it removes the default functionality of opening the emote card.

ddenniem: True, will need to find a way to prevent emote card behaviour when this option is on.

src/Global/EmoteStore.tsx
ddenniem: fixed

Excellify:
I think this is a fine addition and seems to be working. I have a few improvment ideas if you want to do some more development, but not needed if you don't to. 
You could make it so that:

- clicking normaly opens the card/nothing for 3rd party emotes.
- Holding SHIFT adds it to the draft 
- Holding CTRL opends the providers address of the emote (i.e. https://7tv.app/emotes/60a1babb3c3362f9a4b8b33a)

ddenniem:
> I think this is a fine addition and seems to be working. I have a few improvment ideas if you want to do some more development, but not needed if you don't want to. You could make it so that:
> 
>     * Clicking normaly opens the card/nothing for 3rd party emotes.
> 
>     * Holding SHIFT adds it to the draft
> 
>     * Holding CTRL opends the providers address of the emote (i.e. https://7tv.app/emotes/60a1babb3c3362f9a4b8b33a)

I agree, clicking normally should always keep default twitch behaviour to keep it future proof, in case twitch adds a feature when you click an emote. I changed it to shift+click.

I will leave the CTRL click for another feature request.

Excellify (approved):
Looks fine to me

ddenniem:
lmk if a changelog entry should be added

ddenniem:
it should be noted that if ffz is installed, they override this behaviour. Shift+clicking any third party emote with ffz will open a new tab with the emote url

Excellify:
> it should be noted that if ffz is installed, they override this behaviour. Shift+clicking any third party emote with ffz will open a new tab with the emote url

The 7TV extension lets ffz handle rendering of chat lines when both are installed so we cant really controll what happens when emotes are clicked.

AnatoleAM:
What is/are the use case(s) for this?

ddenniem:
> What is/are the use case(s) for this?  

It allows for easier chat copy/pasting. If you see someone use an emote, you can instantly draft it yourself. 
preview
Especially handy if you see the chat spam a specific emote, you can join easily.

PR #218, opened 2022-02-12
Humanized timeout duration

KeterSCP:
This PR adds humanization for timeout messages in chat.

Before:

After:


package.json
```diff
@@ -60,6 +61,7 @@
     "@material-ui/core": "^5.0.0-beta.2",
     "@material-ui/icons": "^5.0.0-beta.1",
     "copy-webpack-plugin": "^7.0.0",
+    "humanize-duration": "^3.27.1",
```
AnatoleAM, line 64: I'd prefer using `date-fns` instead of this package.
See: https://date-fns.org/v2.28.0/docs/intervalToDuration

AnatoleAM (changes requested):

package.json
KeterSCP: Done

src/Sites/twitch.tv/Runtime/ChatListener.ts
```diff
@@ -86,7 +86,7 @@ export class TwitchChatListener {
 			const modMsg = msg as unknown as Twitch.ChatMessage.ModerationMessage;
 
 			if (modMsg.moderationType === 1) { // Timeout
-				this.sendSystemMessage(`${modMsg.userLogin} was timed out for ${humanizeDuration(modMsg.duration * 1_000)}${!!modMsg.reason ? ` (${modMsg.reason})` : ''} by a moderator`);
+				this.sendSystemMessage(`${modMsg.userLogin} was timed out for ${formatDuration(intervalToDuration({start: 0, end: modMsg.duration * 1_000}))}${!!modMsg.reason ? ` (${modMsg.reason})` : ''} by a moderator`);
```
AnatoleAM, line 89: Is there a reason for the underscore at `modMsg.duration * 1_000`?
Also, could you call the `intervalDuration `as a const above rather than inline?

KeterSCP: - The reason for underscore is to determine a thousand parts. For example it's more readable when you see 1_000_000 instead of 100000. I can change it to be 1000 if you wish.
- Sure I'll move it to const

KeterSCP: @AnatoleAM done

Issue #215, opened 2022-01-29
[Bug] Broken emoji width in YouTube comments


davidstraka2:
**Describe the bug**
Emojis in the YouTube comment section are deformed (of unequal height and width). The 7tv extension overrides the default height, but doesn't override the default width. Everything works correctly with the 7tv extension disabled.

**To Reproduce**
1. Go to any YouTube video
3. Scroll down to its comment section and find any comment with emojis
4. See deformed emojis

**Expected behavior**
The emojis should be of equal width & height (or at least most basic emojis).

**Screenshots**
Example of broken/deformed emojis in the comment section:

![broken-emojis](https://user-images.githubusercontent.com/32021723/151669586-bc10cf5f-596f-459f-b9af-d3aee92167e4.png)

The styles causing the issue (Style.css is added by 7tv):

![broken-styles](https://user-images.githubusercontent.com/32021723/151669647-f2d7960c-b826-4ce9-bfe7-b5b090c58534.png)

Temporary fix by adding a custom override (adding the width property):

![fix](https://user-images.githubusercontent.com/32021723/151669679-ca572708-bb00-4a42-878d-09333a795c4f.png)

**Version:**
2.1.4 (though I've had this bug for a long time, only now I've finally got myself to find the cause)

Cryptizism:
Can you close this please @AnatoleAM

PR #220, opened 2022-02-20
Fixed emoji width being altered on YT

Cryptizism:
Emoji width on youtube was sometimes altered by higher priority CSS which used exact pixels of 16 and made the emojis look messed up and thin

#215


Cryptizism:
Made commit on main fork meaning I couldn't add more changes, closing and will re-open with some other CSS changes later

Issue #147, opened 2021-09-24
Emote panel is not showing in Creator Dashboard [Bug] 


YourJest:
**Describe the bug**
Emote panel is not showing properly in Creator Dashboard in Firefox

**To Reproduce**
1. Go to Creator Dashboard
2. Click on 7tv icon near chat

**Expected behavior**
Emote panel is showing properly in Creator Dashboard.

**Version:**
2.0.0
Firefox version 92.0.1
Chrome version 93.0.4577.82

Probably it's because the emote panel showing at the wrong place. When I open developer tools and moving it to the top it appears.
![изображение](https://user-images.githubusercontent.com/16428691/134709619-3e567655-2c5f-4618-a803-66dd4fe81b50.png)

PR #222, opened 2022-02-21
Updated position CSS of the overlay class.

Cameron2920:
Fix for #147 

The parent element of the overlay element differ between https://dashboard.twitch.tv/ and https://www.twitch.tv/. This resulted in the positioning behaviour of the overlay differing on the 2 sites. This change should make sure the position of the overlay is consistent on the 2 sites.


AnatoleAM (approved):
Thank you 👍

Issue #185, opened 2021-11-24
[Bug] Tab-complete for 7tv emotes breaks after being timed out in a chat


zuned11:
After being timed out, tab-complete for 7TV emotes (only 7tv) breaks and stops working. Fixes itself upon refresh.

This is on 2.0.2.

Excellify:
Duplicate of #165

Issue #165, opened 2021-10-13
[Bug] Emote auto-complete stops working after getting timed out


Munzu:
**Describe the bug**
Auto-complete stops working (correctly) after getting timed out. To fix it, I have to refresh the page.

**To Reproduce**
1. Get timed out, no matter the duration
2. Wait until timeout is over
3. Try to auto-complete emote, e.g. `ome`+`Tab` for `OMEGALUL`

**Expected behavior**
`ome` should auto-complete to `OMEGALUL` just like before the timeout

**Version:**
v2.0.0 on Firefox 93.0 (latest stable)

PR #223, opened 2022-02-21
Fixed tab autocomplete not working after a timeout

Cameron2920:
Fix for #165 and #185 

When timed-out, Twitch removes the text input field and adds a new text input field when the timeout expires. This meant that the event listeners in TabCompleteDetection were not listening to the visible text input after a timeout. To handle this, I setup event listeners for each new chat input field that is added (and clear event listeners on old chat input fields).


src/Sites/app/Runtime/TabCompleteDetection.tsx
```diff
@@ -96,15 +92,16 @@ export class TabCompleteDetection {
...
-		input.addEventListener('change', this.finalizeListener, {
-			capture: false
-		});
+		let chatRoomContent = document.querySelector(Twitch.Selectors.ChatRoomContent);
 
+		if(chatRoomContent) {
+			const observer = new MutationObserver(() => {
+				this.updateInput();
+			});
+			observer.observe(chatRoomContent, {subtree: false, childList: true});
```
AnatoleAM, line 102: Mutation observers can be pretty expensive. I believe there should be a way to hook into Twitch's react to detect the timeout states

AnatoleAM (changes requested):

src/Sites/app/Runtime/TabCompleteDetection.tsx
Cameron2920: Thanks, I'll investigate further.

Cameron2920: Never mind the previous comment. I found the ban status here `ChatComponent.props.currentUserBannedStatusData?.channel?.self?.banStatus?.expiresInMs` and added a check for update every 1s.

AnatoleAM: This isn't really what I'm looking for either, unfortunately I am busy with the ongoing work on API V3 and the new website so i cannot guide you on this - but the ideal sitution would be to hook into the exact states that cause the timeout screen to appear/disappear.

Cryptizism: When you get timedout you get sent a gql request with the operation name `ChatRestrictions`

The only way I see doing this without checking `ChatComponent.props.currentUserBannedStatusData?.channel?.self?.banStatus?.expiresInMs` constantly is by [intercepting these request bodys](https://medium.com/@ddamico.125/intercepting-network-response-body-with-a-chrome-extension-b5b9f2ef9466) In this request `[0].data.channel.self.banStatus.isPermanent` will say whether or not the ban is permanent (if it's a timeout or ban). If this is you can check `ChatComponent.props.currentUserBannedStatusData?.channel?.self?.banStatus?.expiresInMs` since it for some reason, from what I can see doesn't include an expiry (weird).From there you can approach it however you'd like.

Snooping around twitch uses, to an extent, the way Cameron said except they don't check every second (obviously) and it's called from somewhere somehow idk, but I'm guessing it's something to do with the GQL request.

Cameron2920: Hey @AnatoleAM I think I've figured out what you're looking for. This should only trigger when the chat component is updated because of a change to the ban status prop (which removes/adds the chat input).

I understand this is probably a low priority issue. If there is something higher priority that I can work on, I am more than happy to help! (I'm fimiliar with front and back end code)

AnatoleAM: That looks better, I'll have a look and test it when I can. Thank you for the fast iterations

src/Sites/app/Runtime/TabCompleteDetection.tsx
```diff
@@ -194,6 +188,86 @@ export class TabCompleteDetection {
 			this.inputValue.next({ message: newMessage, cursorPosition: newCursorPosition });
 		}, 0);
 	}
+
+	/**
+	 * Removes event listeners from the currentInput.
+	 * Finds and sets the new currentInput by querying the DOM.
+	 * Adds event listeners to the new currentInput.
+	 */
+	private updateInput():void{
```
AnatoleAM, line 197: All the methods here could be moved to a more general place, such as InputManager. Tab completion is only one of the elements that becomes broken after being timed out

AnatoleAM (changes requested):
I feel like this could include more than just fixing tab completion, like re-instating the emote menu button as well as history navigation and the duplicate message bypass option.

src/Sites/app/Runtime/TabCompleteDetection.tsx
```diff
@@ -194,6 +188,86 @@ export class TabCompleteDetection {
...
+	 * Listen for chat component updates to the current user's ban status, and update the input on ban status updates.
+	 */
+	private startListeningForBanStatusChanges():void{
+		let chat = (window as any).twitch.getChat();
+		this.originalChatComponentDidUpdate = chat.componentDidUpdate;
+		chat.componentDidUpdate = (previousProps: ChatComponentProps, previousState:ChatComponentState, snapshot:any) => {
+			if(previousProps.currentUserBannedStatusData?.channel?.self?.banStatus != chat.props.currentUserBannedStatusData?.channel?.self?.banStatus){
+				setTimeout(() => this.updateInput(), 0);
+
+				if (!!this.originalChatComponentDidUpdate && typeof this.originalChatComponentDidUpdate === 'function') {
```
Excellify, line 252: Shouldn't the execution of componentDidUpdate be passed on regardless of if the banStatus has changed?

src/Sites/app/Runtime/TabCompleteDetection.tsx
Excellify: The emote menu button needs to be reinjected. Does the "allow sending same message twice" and ctrl + enter also break?

Cameron2920:
> I feel like this could include more than just fixing tab completion, like re-instating the emote menu button as well as history navigation and the duplicate message bypass option.

I've made the ban status checking more general. A ban status change should notify the TwitchPageScript which can then refresh anything that needs refreshing. I've tested to make sure tab complete, chat history, the duplicate message setting and the emote menu are working after becoming untimed out. If there are any features I missed, I can address them.

src/Sites/app/Runtime/TabCompleteDetection.tsx
Cameron2920: Yes, I've updated this.

Cameron2920:
@AnatoleAM thanks for letting me know. If there is anything I can do for 3.0.0, I'd be happy to help.

AnatoleAM:

PR #225, opened 2022-02-23
Hide User-Specific FFZ Emotes

### LINKED ISSUE #212: [Feature request] Handle FFZ User-Specific Global Emotes
The 7TV App shows the "OBOY", "OiMinna", and "AndKnuckles" emotes in the emote menu, which are user-specific global emotes in the "Sten's Cheaty Emotes" list, only usable by sirstendec.

The correct action is to parse the `default_sets` array and add sets with an ID matching that array (currently only set number 3) to the allow list, then go through the `users` associative array and allow access to the set marked by the keys in that array only if the user's name is listed in that key's value array (currently only set number 4330 is listed, with only "sirstendec" in the value array). If the current user is not in the `users` array for a set, that set should not show up in the menu.

The purpose is to let certain users have access to a set of personal emotes in all channels, without asking for the user name with each `/set/global` call.

For direct-source details, see the `/set/global` section of https://www.frankerfacez.com/developers.

Tested on v2.1.4

RealityRipple:
Partially resolves #212

If you don't want to support user-specific global emotes, this is the bare minimum to hide them from the list entirely.

### DIFF
--- a/src/Global/API.ts
+++ b/src/Global/API.ts
@@ -204,6 +204,7 @@ export namespace API {
 	export type EmoteProviderList = [EmoteProvider?, EmoteProvider?];
 	export namespace FFZ {
 		export interface RoomResponse {
+			default_sets: string[],
 			sets: {
 				[key: string]: {
 					emoticons: FFZ.Emote[];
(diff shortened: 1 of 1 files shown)


AnatoleAM (changes requested):
The [type errors](https://github.com/SevenTV/SevenTV/runs/5336815471?check_suite_focus=true) must be resolved

RealityRipple:
Slight note about the type definition: technically the `default_sets` array is made of integers, but they're compared to the `sets` keys, which are strings, and in case IDs ever change style (such as GUID/base64 of some value), it won't need changing.

AnatoleAM:

Issue #221, opened 2022-02-21
[Bug] Incorrect behaviour of moderation slider when user marked by mod as Suspecious is chatting 


marczeslaw11:
**Describe the bug**
Once user marked as Suspecious types single message, the moderation slider the height of entire chatbox appears. Probably caused by miscalculation of the height of message itself. The big slider disappears once chat scrolls enough to completely deload Suspecious User's messages.

**To Reproduce**
1. Mark user as Suspecious on the channel you moderate
2. Have them post a message
3. You can see the moderation slider bugging on the side - it's displayed on entire height of chatbox, with just a tiny part being clickable

**Expected behavior**
The moderation slider should have the same height as the message, and be possible to drag on the entire height of suspecious user's message

**Screenshots**
![image1](https://user-images.githubusercontent.com/49101610/155005098-bbb54ad9-d460-4fe5-80bb-689613c330d9.png)
![image2](https://user-images.githubusercontent.com/49101610/155005688-73f64244-20ca-4f6a-9513-b879db588724.png)

**Version:**
2.1.4 for Mozilla Firefox

Cameron2920:
The ban slider for suspicious messages isn't displayed correctly because of styling differences between a suspicious message and a normal message. I can probably make a PR to handle this.

marczeslaw11:
Would be appreciated if you did

marczeslaw11:
thanks for help

PR #226, opened 2022-02-25
Updated ban slider styling to support suspicious user messages.

Cameron2920:
Fix for #221 

The chat line for a suspicious message has position static, and the ban slider is laid out assuming the chat line has position relative. This change will make all chat lines with a ban slider position relative. If you think this is unsafe, I can restrict this change to apply only for suspicious message chat lines. Alternatively, I can modify the ban slider to work with a position static chat line.

### DIFF
--- a/src/Sites/twitch.tv/Components/BanSliderManager.tsx
+++ b/src/Sites/twitch.tv/Components/BanSliderManager.tsx
@@ -84,6 +84,7 @@ export class BanSliderManager {
 
 					const container = document.createElement('span');
 					container.classList.add('seventv-ban-slider');
+					line.element.style.position = 'relative';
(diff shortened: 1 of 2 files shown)


src/Sites/twitch.tv/Components/BanSliderManager.tsx
```diff
@@ -84,6 +84,7 @@ export class BanSliderManager {
 
 					const container = document.createElement('span');
 					container.classList.add('seventv-ban-slider');
+					line.element.style.position = 'relative';
```
AnatoleAM, line 87: I'm not a huge fan of changing the entire chat line into relative here

AnatoleAM (changes requested):

Issue #197, opened 2022-01-06
The logo 7tv uses for settings and emote menu's is blurry.


treuks:
**Describe the bug**
The logo 7tv uses for settings, and the emote picker, is blurry, due to it being scaled down from a **512x512** **webp** image

**To Reproduce**
1. Go to a twitch page, or a youtube live stream
2. Look at the 7tv logo
3. See that it is blurry

**Expected behavior**
It should be a nice and not blurry 7tv logo

**Screenshots**
![image](https://user-images.githubusercontent.com/18615981/148410647-2d937db8-6a99-4840-b186-b63c0140dde6.png)

![image](https://user-images.githubusercontent.com/18615981/148410695-a6f40188-bdfb-4328-b3cc-4653426767d6.png)

![image](https://user-images.githubusercontent.com/18615981/148410741-9d02fd79-b1fa-4980-824e-839cd15b4b04.png)

**Version:**
Chromium 94.0.4606.61
SevenTV 2.0.2

NS2555:
As a solution, i have created a nice and pixel aligned svg version of the logo 

```
<svg width="16" height="16" viewBox="0 0 4.233 4.233" xmlns="http://www.w3.org/2000/svg"><path style="fill:#0e0e10;fill-opacity:1;stroke:none;stroke-width:.264583px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1" d="m21.431 7.408-1.058 1.059-1.323-1.323V8.73l-.53-.529V6.35l1.588-1.058-2.645.264-.265-.529 4.233-.53v.53l-2.116 1.588 1.058 1.058 1.058-1.058z" transform="translate(-17.198 -4.498)"/></svg>
```
You are free to use it everywhere for the extension

Excellify:
It dosn't appear blurry to me?

NS2555:
look at how sharp the edges are on the native website buttons compared to the 7tv button

Excellify:
Hmm, I can maybe see some blur on the images you linked. I dont have the same blur when its rendered for me tho. :thinking: 
What resolution is this?

NS2555:
1920x1080p chromium on linux

NS2555:
hmm, it seems like it looks better in firefox 
![image](https://user-images.githubusercontent.com/18615981/148424650-b9086c77-403a-4a30-892e-3ec120267c06.png) (firefox)

[...]

PR #236, opened 2022-03-25
Fixed 2 minor CSS issues

Cryptizism:
(Making a new PR for this as last PR I committed to my repo and the dev branch so I couldn't make additional commits/edits.)(Previous PR: https://github.com/SevenTV/SevenTV/pull/220)

## #215 - Emoji width messed up

https://github.com/SevenTV/SevenTV/commit/c732e5458814dd5900501a4d68c9202e3efd337c prevents other CSS with higher priority from altering the image width since height is !important the emoji is unable to adapt to the higher priority CCS and just takes the height with it

image

## #197 - Blurry logo

https://github.com/SevenTV/SevenTV/commit/be7aa89f99370c11990aa710724a566f6fc60ff0 fixes the logo being slightly blurry on the site

With changes:
image
Prior to changes:
image

### DIFF
--- a/src/Style/Style.scss
+++ b/src/Style/Style.scss
@@ -78,6 +78,7 @@ div[id="content"]
 
 .emoji {
 	height: 1.95em !important;
+	width: auto !important;
 }
 
 // Move the YouTube Slowmode countdown back over the Send button
--- a/src/Style/Style.scss
+++ b/src/Style/Style.scss
@@ -219,7 +219,6 @@ div[id="content"]
   		width: 100%;
   		height: 100%;
 		mask-size: contain;
-		image-rendering: -webkit-optimize-contrast;
   		-webkit-mask-size: contain;
 	}
 }
--- a/src/Style/Style.scss
+++ b/src/Style/Style.scss
@@ -219,6 +219,7 @@ div[id="content"]
   		width: 100%;
   		height: 100%;
 		mask-size: contain;
+		image-rendering: -webkit-optimize-contrast;
   		-webkit-mask-size: contain;
 	}
 }


Cryptizism:
No idea why the PR build is failing, it's a typescript error and no errors occur on my PC/local version

PR Build:
image

Locally:
image

Plus it's just CSS idk why it would cause errors

src/Style/Style.scss
```diff
@@ -218,6 +219,7 @@ div[id="content"]
   		width: 100%;
   		height: 100%;
 		mask-size: contain;
+		image-rendering: -webkit-optimize-contrast;
```
AnatoleAM, line 222: What is this for? I'm not able to find any documentation on this property, which means it may be unsupported

AnatoleAM (changes requested):

src/Style/Style.scss
Cryptizism: > What is this for? I'm not able to find any documentation on this property, which means it may be unsupported

With changes:
image
Prior to changes:
image

In essence, optimized contrast is a rendering mode that gives "crisp edges"

The property works on: Chrome, Edge, Opera, and Safari. (https://www.w3schools.com/cssref/tryit.asp?filename=trycss3_image-rendering)

When on Mozilla firefox I believe the image rendering is automatically different as the logo does not appear to be blurry as seen below (without changed image rendering.)
Mozilla:
image
Chrome:
image

Overall, the change is extremely minute and is not that integral. You could also probably make use of an SVG instead to prevent blurring.

AnatoleAM: Can you remove this? The other change seems fine, but I don't like this change. The icon could be changed to a pre-scaled image or a vector, applying this property doesn't seem to make it look better; just more pixelated.

Cryptizism: Reverted 5986729d65b23277ede8d86e620681090f839f8e

Cryptizism:
@AnatoleAM How is it now? I reverted the logo CSS change

AnatoleAM (approved):

PR #258, opened 2022-05-04
Fix video cutoff in true theater mode on youtube

Excellify:
Youtube must have made the video player scale to cover instead of contain recently. This caused the video to be cut off when using the true theater mode.


AnatoleAM (changes requested):
Can you add to changelog.md under version 2.2.2

AnatoleAM (approved):

PR #259, opened 2022-05-06
Update gitmodules to support both https and ssh

Excellify:
Allows for cloning the repository with both https and ssh. 
The current config causes errors if the user has not set up ssh keys.


AnatoleAM:

PR #268, opened 2022-05-28
Fix rendering of 3rd party emotes in : autocomplete tray

Excellify:
Twitch recently changed how they render the emotes in the autocomplete tray. The renderEmoteSuggestion is no longer called, so we need to override the srcSet elsewhere. Seems to be working when overriding the getMatches.


AnatoleAM (changes requested):
This appears to remove the provider indicators in colon completion. (Perhaps because the code block which handled this has been removed for some reason?)

Excellify:
> This appears to remove the provider indicators in colon completion. (Perhaps because the code block which handled this has been removed for some reason?)

The best solution i could find was editing the srcSet in the getMatches function. And there i only had access to the react element. I dont know what the provider indicators are supposed to do, but couldnt find anything to "update" in the element accordingly.

AnatoleAM:
It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`

Excellify:
> It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`

Does twitch do that by default? We could alternatively just add edit the displayed text

AnatoleAM:
> > It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`
> 
> Does twitch do that by default? 

No. This is something I added, you can see it in the code block you removed in this PR

Excellify:
> > It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`
> 
> Does twitch do that by default? We could alternatively just add edit the displayed text

Seems like `provider` isnt available in the scope of this function making it difficult to add it to the end. We could get the id and then find it by iteration over all of the emote sets, but dont think that would be a good solution

AnatoleAM:
> > > It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`
> > 
> > 
> > Does twitch do that by default? We could alternatively just add edit the displayed text
> 
> Seems like `provider` isnt available in the scope of this function

That's because it was added to the emote objects. Remember all objects in JavaScript are pointers.

Excellify:
> > > > It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`
> > > 
> > > 
> > > Does twitch do that by default? We could alternatively just add edit the displayed text
> > 
> > 
> > Seems like `provider` isnt available in the scope of this function
> 
> That's because it was added to the emote objects. Remember all objects in JavaScript are pointers.

Yea, was just stating that we dont have access the provider of the emote element that is returned to us. So that we'd have to resolve the emote again from using the id if we want to find the provider.

AnatoleAM:
> > > > > It added information about where any given emote came from, i.e `Kappa (Twitch Global)` or `GIGACHAD (Channel 7TV)`
> > > > 
> > > > 
> > > > Does twitch do that by default? We could alternatively just add edit the displayed text
> > > 
> > > 
> > > Seems like `provider` isnt available in the scope of this function
> > 
> > 
> > That's because it was added to the emote objects. Remember all objects in JavaScript are pointers.
> 
> Yea, was just stating that we dont have access the provider of the emote element that is returned to us. So that we'd have to resolve the emote again from using the id if we want to find the provider.

Oh, so Twitch changed something. Then yeah I guess that's the only way to reinstate this feature

AnatoleAM (approved):

Issue #230, opened 2022-03-10
recent-messages


albertopasqualetto:
Please add support to recent-messages.robotty.de, it is an interesting feature.

PR #279, opened 2022-07-17
Add support for message history using robotty/recent-messages API

tarunbod:
## Description

Implements #230. Adds:
- Three new settings for the user to configure
  - "Show Recent Chat History" - checkbox, enables/disables the feature.
  - "Limit Recent Chat History" - radio select, max number of historical chat messages to pull from the recent-messages API
  - "Recent Chat History Time Limit" - radio select, how far back in time to pull messages from
- Loads as many historical messages as are present in the configured time limit, up to the max configured number of messages
- Shows historical messages in the chat once the chat listener is loaded, with timestamps

## Considerations
- Thought about the traffic we'd be sending to the recent messages API and whether it's a good idea, but since chatterino seems to use it and there's no specified rate limits, I think it should be fine. Current implementation only calls the API once per page load (or reload).
- Decided to get and show the recent messages after the chat listener is started so it will intercept those messages and perform third party emote rendering. As is, the PR doesn't auto-render twitch emotes in historical messages, could add that if wanted
- Made the entire feature opt-in due to robotty/recent-messages' disclaimer requirement: `https://recent-messages.robotty.de/api`
- The API URL is hardcoded, wonder if there's some config file where it would fit better?


src/Sites/twitch.tv/Runtime/RecentMessagesProvider.ts
```diff
@@ -0,0 +1,98 @@
+import { Twitch } from 'src/Sites/twitch.tv/Util/Twitch';
+
+const RECENT_MESSAGES_API_URL = 'https://recent-messages.robotty.de/api/v2';
+
+export class RecentMessagesProvider {
```
tarunbod, line 5: As it stands only actual chat messages are supported. The API also returns messages like sub notifications, but I based this PR on how historical messages are loaded when you're the moderator of a channel. The chat history that mods see on page load doesn't include those kinds of messages, so I didn't bother with adding support for that. But I don't think it would be too difficult to do so.

Excellify:
Dosnt twitch have an experiment currently running that enables recent messages for everyone, not just mods?

tarunbod:
Yes, I didn't know about this prior to opening this PR. FFZ experiments tab says that it's currently being shown to 5% of users. Do you know how often experiments get released as full features? If it's common then I don't think this PR has much point and I'll probably close it out.

PR #287, opened 2022-08-13
Temporary workaround to messages sendt by the user being rerendered.

Excellify:
This fixes the rerendering of messages send by the user. Not sure of the implications of not clearing the patcher, but dosn't seem to break anything immediately. Not including changelog in case you want to improve upon this. Ill be gone for most of the rest of the day.


AnatoleAM:
Seems to be fine, I am not sure about memory implications of this, but it should get garbage collected when the message is destroyed I suppose.

squareSphere29:
It's not completely fixed, it is still occuring

Excellify:
> It's not completely fixed, it is still occurring

In what situation does it still happen? Note that this version is still not on the chrome store.

PR #355, opened 2023-03-23
Add highlight style options

Melonify:


AnatoleAM (approved):
this is the greatest highlight style options dropdown of all time

PR #367, opened 2023-03-23
Add sidebar-tooltip stream previews

Melonify:
Adds option to show thumbnail previews of streams when hovering over channels on the sidebar


8thony:
For me its to small I want to have the option to increase this preview and the image quality.
I quickly tested this by setting the image quality to 1920x1080 and the box width to 100.
Note: if the browser window gets smaller, it should automatically decrease the preview box size otherwise it fills the entire window

MaskedDuckling311:
a thumbnail size slider and a hover delay option would be great. 
otherwise, the implementation looks clean.

AnatoleAM (approved):

PR #369, opened 2023-03-24
Don't parse fake emotes patched into messages by BTTV and FFZ

Melonify:
Fixes issue causing emotes to sometimes display as broken images


AnatoleAM (approved):
this is the greatest anti-internals poisoning measure of all time

PR #414, opened 2023-03-29
Fix unwanted nametags paint tooltip

### LINKED ISSUE #410: [Bug] Nametag Paint-names shown always
if setting "Nametag Paints" is turned off and you hover a user that has a Paint active the name of this Paint is displayed

MatinAniss:
Added check that removes the nametag paint name tooltip if nametag paints are disabled.

BUG screenshot:
Screenshot_4

FIXED screenshot:
Screenshot_5

Fixes #410


Melonify (changes requested):
Not sure this is the right way to do this, I think it would be preferable to instead not set `paint` when the setting is disabled.

AnatoleAM (changes requested):
`paint` should simply not be set when the option is disabled, rather than adding this extra check in the template.

MatinAniss:
ah, that is what you meant. I changed how name tags paints should be disabled overall which removes the old method which was commented to be replaced. This way the nametag paint and tooltip are disabled without extra checks like I did. :)

AnatoleAM (approved):

PR #435, opened 2023-03-30
feat:add seconds to timestamp in twitch chat

### LINKED ISSUE #420: Add seconds to timestamps in chat
Requested by @CrazyWolf13 in #411

TheRealGI:
Close #420

### DIFF
--- a/src/site/twitch.tv/modules/chat/components/message/UserMessage.vue
+++ b/src/site/twitch.tv/modules/chat/components/message/UserMessage.vue
@@ -232,16 +232,18 @@ watchEffect(() => {
 		);
 	}
 
-	timestamp.value = intlFormat(
-		{ locale },
-		{
-			localeMatcher: "lookup",
-			hour: "numeric",
-			minute: "numeric",
-			second: displaySecondsInTimestamp.value ? "numeric" : undefined,
-		},
-		props.msg.timestamp,
-	).replace(/ (A|P)M/, "");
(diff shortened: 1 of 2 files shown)


src/site/twitch.tv/modules/chat/components/message/UserMessage.vue
```diff
@@ -153,6 +153,7 @@ const timestamp = intlFormat(
 		localeMatcher: "lookup",
 		hour: "numeric",
 		minute: "numeric",
+		second: "numeric",
```
AnatoleAM, line 156: This would make it permanently show seconds, which is not a default behavior. It must be togglable, and off by default

AnatoleAM (changes requested):

src/site/twitch.tv/modules/chat/components/message/UserMessage.vue
TheRealGI: like adding a new option in the 7tv settings at Chat > Style ?

robmonte: @TheRealGI I expect it makes the most sense to put it there. I'm looking forward to having seconds. 😄

AnatoleAM (approved):
Made some changes to wording, and made the timestamp reactive so that it updates on existing messages when the setting is toggled.

Thank you for your contribution.

PR #457, opened 2023-04-04
Add a tooltip to show full "Replying to @..." text

tautautautautau:
A tooltip to view the entire message being replied to, eliminating the need to open the thread to read it.


Melonify:
This looks fine, but it could be drastically simplified using the `v-tooltip` custom directive.

Just setting `v-tooltip` on any element will setup a tooltip exactly like the one you've created manually

AnatoleAM (commented):
Not sure I like how it appears in the middle :thinking: looks a bit goofy when in popout mode

tautautautautau:
@AnatoleAM This PR is ready. Could this be merged since it is a QoL that some users have asked for?

PR #459, opened 2023-04-05
fix: bad ci condition syntax

AnatoleAM:
woops


AnatoleAM:
this is why you don't push straight to master 😤

PR #465, opened 2023-04-05
Fix setting controls not scrolling with the rest of the settings menu

IZEDx:
My twitch has an inline style tag that sets `.control { position: absolute; } `, which results in the settings menu being unusuable.

I thought it was a compatibility issue as there's no open Issue about that, so I disabled all other twitch extensions and that didn't fix the issue. This is quite a serious issue and since nobody else reported it I guess it's an edge case with my system.

Anyway, this fix is just a single line of css that shouldn't conflict with anything else. I didn't bother to create an open issue for it, I hope that's not a problem.

For reference, this is what it looks like when I scroll down in the settings:

image
image

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -1,7 +1,5 @@
 ### Version 3.0.4.7000
 
-**The changes listed here are not assigned to an official release**.
-
 -   Added hot-patching functionality to the extension
 -   Added chat rich embeds which allows twitch clips to preview in chat
 -   Added a "Copy Message" button
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -1,4 +1,4 @@
-### Untitled Version
+### Version 3.0.4.7000
 
 **The changes listed here are not assigned to an official release**.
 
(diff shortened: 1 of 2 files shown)


AnatoleAM:
Isn't this an extension using the same `.control` class, thereby adding extra styling? If so, it would be better to simply add a prefix to the classes here. This is something we do everywhere else, i.e `seventv-settings-...`

IZEDx:
> Isn't this an extension using the same `.control` class, thereby adding extra styling? If so, it would be better to simply add a prefix to the classes here. This is something we do everywhere else, i.e `seventv-settings-...`

You are right, it'd be better to add a prefix here. I didn't question it, as it's a nested class in scss, but since there's no shadow-dom it'd make sense to future proof all generic class names being used by enforcing prefixes. Question is, does it make sense to make that change in this PR already or leave it as it is, considering this affects many more components and a new Issue has to be created anyway?

Melonify (changes requested):
Yeah, I think prefixing is the way to go here, this is an incompatibility with the Twitch DVR player extension you have enabled.

Setting position to unset is not a good fix here.

IZEDx:
> Yeah, I think prefixing is the way to go here, this is an incompatibility with the Twitch DVR player extension you have enabled.
> 
> Setting position to unset is not a good fix here.

Good to know it's an incompatibility issue with Twitch DVR, must've missed it when going through my extensions.

I changed it to use a prefix instead of using position unset. I only added a prefix to the control class however, the other classes should be prefixed in a separate branch for futureproofing. There are quite a lot of classes that have not been prefixed yet scattered around the whole project, which could collide with other extensions in the future.

Melonify (approved):

PR #479, opened 2023-04-08
Firefox Self-Distribution

AnatoleAM:
Since addons.mozilla.org is an absolute disaster, we are switching to self-distribution on Firefox.


AnatoleAM:
Currently stalled due to https://github.com/mozilla/addons/issues/1481

PR #489, opened 2023-04-10
fix: decimal no longer rendered as link

### LINKED ISSUE #488: Decimal are rendered as link in Twitch chat
image
Decimal should not be displayed as link

TheRealGI:
Closes #488

Changed the link regex to only match messages which starts with http, https or www. 
It won't be possible to write google.com in chat which was previously rendered as https://google.com

### DIFF
--- a/src/common/chat/Tokenizer.ts
+++ b/src/common/chat/Tokenizer.ts
@@ -8,6 +8,7 @@ import type {
 	VoidToken,
 } from "@/common/chat/ChatMessage";
 import { Regex } from "@/site/twitch.tv";
+import { parse } from "tldts";
 
 const URL_PROTOCOL_REGEXP = /^https?:\/\//i;
 
@@ -74,7 +75,7 @@ export class Tokenizer {
 			} else if (prevEmote && part.startsWith("ffz") && part.length > 3) {
 				// this is a temporary measure to hide ffz emote modifiers
 				tokens.push(toVoid(cursor, next - 1));
-			} else if (part.match(Regex.Link)) {
+			} else if (this.isValidLink(part)) {
 				// Check link
 				const actualURL = part.replace(URL_PROTOCOL_REGEXP, "");
 
@@ -113,6 +114,10 @@ export class Tokenizer {
 		tokens.sort((a, b) => a.range[0] - b.range[0]);
 
 		return (this.msg.tokens = tokens);
+	}
+
+	private isValidLink(message: string): boolean {
+		return parse(message).isIcann ?? false;
 	}
 }
 
--- a/src/common/chat/Tokenizer.ts
+++ b/src/common/chat/Tokenizer.ts
@@ -26,8 +26,7 @@ export class Tokenizer {
 
 		let cursor = -1;
 		let lastEmoteToken: EmoteToken | undefined = undefined;
-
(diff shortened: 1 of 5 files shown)


Melonify:
Not a fan of this change, we should be using actual standards defined pattern matching for URIs.

This patch places us out of line with how other clients handle links, as well as native. Links which are valid but not prefixed or suffixed should be considered as links. 

We may desire to use a TLD based test like the domain listing from Mozilla, this is the behavior of Chatterino at least. This is also achievable by using a small weight library which bundles the tasks of link matching and domain validity.

TheRealGI:
After a bit of research I found this [library ](https://www.npmjs.com/package/tldts)which can parse URL and returns further informations.
Now the current implementation is that the message gets parsed in the library and as a result I get a object which includes a boolean ``` "isIcann" ``` which indicates if the TLD is from the ICANN list. If  ```null```  or ``` false ``` it's not a link. I also removed the link regex as IMO isn't needed anymore.

Melonify:
:thinking: I think this is almost close to being perfect.

My only concern is that the library makes no other concerns with URL semantics than the domain name itself.
I think this can be easily rectified though, I think doing two checks here makes sense.

First check if the word/token matches from `tldts` and is a valid domain, then use the native URL() parsing, or alternatively we could parse with URL first then use `tldts` only on the `hostname` part of the URL.

Something like 
```js
try {
   const url = new URL(word.startsWith(/[a-z]*:/i) ? word : `https://${word}`);
   
   if (url.protocol == "http:" || url.protocol == "https:") {
      const domain = tldts.parse(url.hostname);

      if (domain.isIcann || domain.isIP) {
           return url;
      }
   }
} catch (e) {void 0;}

return null;
```

Checking this way ensures that both the URL is valid, and the domain is valid, loosely permitting the lack of an explicit scheme, but not permitting any schemes which are NOT http, this would be dangerous, and is my primary source of concern with just using a simpler mechanism, we want to be very sure that links won't open external programs or etc.

Then additionally `URL_PROTOCOL_REGEXP` is no longer necessary as we can pass URL.toString directly to the `url` field of the token.

src/common/chat/Tokenizer.ts
```diff
@@ -8,6 +8,7 @@ import type {
 	VoidToken,
 } from "@/common/chat/ChatMessage";
 import { Regex } from "@/site/twitch.tv";
+import { parse } from "tldts";
```
AnatoleAM, line 11: nit: make this a default import

```suggestion
import tldts from "tldts";
```

src/common/chat/Tokenizer.ts
```diff
@@ -114,6 +115,10 @@ export class Tokenizer {
 
 		return (this.msg.tokens = tokens);
 	}
+
+	private isValidLink(message: string): boolean {
+		return parse(message).isIcann ?? false;
```
AnatoleAM, line 120: nit: namespace with default import

```suggestion
		return tldts.parse(message).isIcann ?? false;
```

AnatoleAM (approved):
I'm not quite sure what the purpose of the callback was for, so I went ahead and changed it to a normal return value. I've also aliased the import value as it seems the `tldts` package has a malformed export structure.

Will let this cook in the nightly release for a little to ensure it works as expected; thank you.

Issue #496, opened 2023-04-12
.yarn directory should be added to .gitignore


berghall:
I'll make a PR soon

PR #498, opened 2023-04-12
chore/include-yarn-gitignore: Include .yarn directory in .gitignore

berghall:
Original issue:  #496


AnatoleAM:
Under what condition is this directory generated? And what purpose does it serve?

PR #501, opened 2023-04-13
feat: add an option to highlight messages from moderators

lumynou5:
This adds an option to highlight messages from moderators.
The green (`#00bb00`) for highlights is the same color on the moderator badge.

This is reopen of #486.

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -4,6 +4,7 @@
 
 -   Links in chat messages now respect known TLDs instead of matching any url-like pattern
 -   Added an option to show timeouts/bans directly in the chat without being a moderator
+-   Added an option to highlight messages from moderators.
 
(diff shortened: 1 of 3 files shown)


lumynou5:
Please check it out.

AnatoleAM:
Closing this because moving forward we should be inferring the values from Twitch's own highlight settings, this is too confusing to the end user to be met with 2 identical but clashing systems.

lumynou5:
> Closing this because moving forward we should be inferring the values from Twitch's own highlight settings, this is too confusing to the end user to be met with 2 identical but clashing systems.

Where is Twitch's own settings?

lumynou5:
Thanks.
It seems designed for streamers and moderators, but I want viewers can see highlights. I think we can read Twitch's settings for streamers and mods (sync between them?), and for viewers they can adjust their own.

PR #513, opened 2023-04-15
fix: displayName mentions

### LINKED ISSUE #471: [Bug] [Twitch] International name @ mention event doesn't trigger
I have an international name set and if someone @s me using the international name, it doesn't trigger the mention event, it just shows the message normally. If I get an @ using my english twitch name, everything works as expected.

lukashoracek:
This adds/fixes support for mentions using the displayName rather than the user's international username

JavaScript RegExp does not support unicode `\b` or `(*UCP)` so I had to use this ugly workaround, which is used only if the user has different display name than username. I am not sure if there is any better solution than this. [Lookbehind support](https://caniuse.com/js-regexp-lookbehind) should be fine considering 7TV does not have a Safari extension
I also changed the mention regex to `@\S*` otherwise unicode mentions are not matched

Fixes #471

### DIFF
--- a/src/site/twitch.tv/modules/chat/ChatList.vue
+++ b/src/site/twitch.tv/modules/chat/ChatList.vue
@@ -464,10 +464,15 @@ watch(pageVisibility, (state) => {
 watch(
 	[identity, showMentionHighlights, shouldPlaySoundOnMention, shouldFlashTitleOnHighlight],
 	([identity, enabled, sound, flash]) => {
-		const rxs = identity ? `\\b${identity.username}\\b` : null;
-		if (!rxs) return;
+		if (!identity) return;
 
-		const rx = new RegExp(rxs, "i");
+		// Handle both username and displayName mentions
+		// Workaround \b not working properly with unicode characters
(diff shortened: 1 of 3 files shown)


src/site/twitch.tv/modules/chat/ChatList.vue
```diff
@@ -464,10 +464,15 @@ watch(pageVisibility, (state) => {
...
-		const rx = new RegExp(rxs, "i");
+		if (!identity) return;
+
+		// Handle both username and displayName mentions
+		// Workaround \b not working properly with unicode characters
+		const hasMultiLangName = "displayName" in identity && identity.username !== identity.displayName.toLowerCase();
+		const rxs = hasMultiLangName
+			? `((?<=^|\\P{L})${identity.username}|${identity.displayName}(?=\\P{L}|$))`
+			: `\\b${identity.username}\\b`;
+		const rx = new RegExp(rxs, hasMultiLangName ? "iu" : "i");
```
AnatoleAM, line 475: This seems a little overengineered

Issue #518, opened 2023-04-15
[Request] Setting to hide mod chat highlights


Niller2005:
Would be nice to have an option to change opacity and/or hide the  `Monitored Suspicious User` highlights for mods.

How it looks now:
![image](https://user-images.githubusercontent.com/3404570/232255937-fe62610a-f6f0-4692-b4b0-bda27c0a6690.png)

MatinAniss:
#519 PR merged, issue closed.

PR #519, opened 2023-04-16
feat: add option to hide suspicious user highlight

Niller2005:
#518

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -7,6 +7,8 @@
 -   Added options to change what emotes are displayed in the colon list and tab-completion carousel
 -   Fixed emote cards sometimes not showing who added the emote
 -   Fixed an issue where the detailed emote card would clip under existing chat messages
+-   Added an option to hide suspicious user highlights
+-   Added an option to hide low trust user highlights
 
 ### Version 3.0.5.1000
 
--- a/src/site/twitch.tv/modules/chat/ChatList.vue
+++ b/src/site/twitch.tv/modules/chat/ChatList.vue
@@ -67,6 +67,7 @@ const showMentionHighlights = useConfig("highlights.basic.mention");
 const showFirstTimeChatter = useConfig<boolean>("highlights.basic.first_time_chatter");
 const shouldPlaySoundOnMention = useConfig<boolean>("highlights.basic.mention_sound");
 const shouldFlashTitleOnHighlight = useConfig<boolean>("highlights.basic.mention_title_flash");
+const showSuspiciousUser = useConfig<boolean>("highlights.basic.suspicious_user");
 
 const messageHandler = toRef(props, "messageHandler");
 const list = toRef(props, "list");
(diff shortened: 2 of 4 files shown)


src/site/twitch.tv/modules/chat/ChatList.vue
```diff
@@ -139,7 +140,7 @@ function onChatMessage(msg: ChatMessage, msgData: Twitch.AnyMessage, shouldRende
 		msg.setComponent(typeMap[0], { msgData: msgData });
 	}
 
-	if (msgData.type === MessageType.RESTRICTED_LOW_TRUST_USER_MESSAGE) {
+	if (msgData.type === MessageType.RESTRICTED_LOW_TRUST_USER_MESSAGE && showSuspiciousUser.value) {
```
MatinAniss, line 143: Low Trust User ≠ Suspicious User

In my opinion their should be another setting to enable or disable low trust user highlights.

CHANGELOG-nightly.md
```diff
@@ -4,6 +4,7 @@
 
 -   Links in chat messages now respect known TLDs instead of matching any url-like pattern
 -   Added an option to show timeouts/bans directly in the chat without being a moderator
+-   Add option to hide suspicious user highlight
```
lukashoracek, line 7: This should probably be "Added an option to hide suspicious user highlights" to match with the rest of the changelog

src/site/twitch.tv/modules/chat/ChatModule.vue
```diff
@@ -353,6 +353,18 @@ export const config = [
...
+		label: "Show Suspicious User Highlights",
+		hint: "Whether or not to highlight users who are marked as a suspicious user",
+		defaultValue: true,
+	}),
+	declareConfig<boolean>("highlights.basic.low_trust_user", "TOGGLE", {
+		path: ["Highlights", "Built-In"],
+		label: "Show Low Trust User Highlights",
+		hint: "Whether or not to highlight users who are marked as a low trust user",
+		defaultValue: true,
+	}),
```
AnatoleAM, line 367: "Low Trust User" and "Suspicious User" are synonymous. The former is just the technical term used by Twitch. The only differentiation is "Monitored" and "Restricted".

AnatoleAM (changes requested):

src/site/twitch.tv/modules/chat/ChatModule.vue
Niller2005: Should both of these options be moved to "Chat > Moderation" instead and change the Restricted to opacity so it's always visible?

AnatoleAM: Keep it under highlights for now. Ideally these things should actually sync up with the native highlights menu from twitch, but we don't have hooks into this yet. I may look into it sometime if no one else does

AnatoleAM (approved):
This is fine for now.

I'd like to note though in the future we should look at getting this from twitch's highlight settings.

PR #525, opened 2023-04-16
feat: Channel specific highlights via Channel name

MisterXYCB:
Allows to set a channel id for custom highlights

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,49 @@
+name: 🐛 Bug Report
+description: File a bug report
+title: "[BUG] <title>"
+labels: bug
+body:
+  - type: checkboxes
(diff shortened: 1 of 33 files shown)


AnatoleAM (changes requested):
Channel ID isn't a user-friendly parameter, it's virtually impossible for a regular user to find that value. This would need to either use login name, or make an intermediate query for the channel ID. (see `useApollo` & related code for interacting with Twitch's internal apollo instance)

AnatoleAM:
In general I think this isn't the best approach to take, it would be best to implement a profile system with condition-based activation.

MisterXYCB:
Changed it to work with channel names

MisterXYCB:
@AnatoleAM changed it to use channel name

AnatoleAM:
Closing as this is a wrong approach at implementing a feature like this. It should be part of a wider profile system

Issue #458, opened 2023-04-05
Option to hide Leaderboard above chat


Melonify:

Treadspear:
I have made a post about this in discussion: https://github.com/SevenTV/Extension/discussions/467

consider bumping it for me?

ZemyCode:
https://github.com/SevenTV/Extension/pull/481

Addressed in this PR.

AnatoleAM:
Implemented by #481

PR #481, opened 2023-04-08
Add Ability to Hide Unwanted Page Elements

ZemyCode:
## Hide Elements Feature

This feature addition adds a new menu in the settings for hiding certain page elements based on requests from different issues.

This is what the new menu looks like: 

image

Various settings added with false being default value for all of them:

image
image
image

Feedback will be wanted for the names and descriptions of the settings.

Also requesting feedback on the performance or code quality as I am not sure if calling "updateHiddenElements()" on every setting change is optimal.

**Edited 4/27/2023:** Removed "Home Page" feature for hiding Front Page Carousel due to being buggy, may look into this more in the future.

PR #535, opened 2023-04-23
added a option to hide the sub leaderboard

ZKLaurin:
Added a option to hide the Subscriber Leaderboard on top of chat.
This is in response to #458 and #467


ayyybubu:
Looks like  #481 already covers that

ZKLaurin:
yea it does, also way cleaner. Sorry to bother :)

PR #537, opened 2023-04-24
changed readme to include better windows instructions

ZKLaurin:
Added more clear steps for Windows and also added the License for a quicker overview.


Melonify (changes requested):
This is wrong information, if anything it should just instruct people on how to install GNU Make on Windows, instead of trying to awkwardly sidestep our build chain.

Melonify:
Also the license paragraph is not correct, and I think that entire section should be omitted.

ZKLaurin:
I basically removed all the stuff for windows and just made a step by step guide on how to install the necessary packages for WSL.

ZKLaurin:
I need to close this PR, because i commited into the master branch

PR #540, opened 2023-04-24
feat/timestamp_12hour_format: Added an option to toggle timestamp format

### LINKED ISSUE #527: [Request] Toggle AM/PM for Timestamps
Would it be possible to add an option to toggle the AM/PM for timestamps? or even toggle 7tv's control over timestamps and leave it like the default twitch setting

berghall:
I was thinking of adding a system preference option and detect the cycles by locale, but it felt maybe a bit overkill.

Closes #527


TheRealGI:
Well the format already is applied by locale. For e.g. when I set the preferred language to german I have the 24h format when I set it to english I get 12h format. --> Chrome Settings > language > preferred language

Melonify:
Just omitting the value should inherit from the current locale, so this could be easily accomplished via a dropdown option, Infer (Default), Enabled, Disabled. If the setting is set to infer then just pass `undefined`.

AnatoleAM (approved):
LGTM

PR #542, opened 2023-04-26
feat/message_buttons_color: Added an option to change message buttons color

berghall:
e513dee443cffd7e86a50bbb41578abc


Melonify (changes requested):
`v-bind` should be avoided here, it causes inexplicable performance issues during reconciliation. A better approach may be to just set a custom CSS prop on the body tag, in the `effect` of the settings node, this is how many of style the settings currently operate.

berghall:
@Melonify your requested changes have been implemented.

AnatoleAM:
Not exactly a desirable design change

PR #544, opened 2023-04-27
Automatically close emote menu after selection

DanielFirpo:
It's annoying to have do an extra click to close the emote menu after making your selection, this change automatically closes it after clicking an emote.

If the user wants to select multiple emotes, they can hold shift while clicking. This is the same way BTTV menu works.

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,7 +2,8 @@
 
 **The changes listed here are not assigned to an official release**.
 
--   Automatically close emote menu after selection, unless shift key is held down
+-   Automatically close the emote menu after sending a chat message
+-   Added option to automatically close emote menu after making a selection, unless the shift key is held down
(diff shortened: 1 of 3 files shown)


Melonify (changes requested):
Changing set precedent behavior is not desirable, this should be behind a settings option, defaulted to the previous state.

DanielFirpo:
@Melonify I implemented that requested change. I put the setting under appearance even though it's a functionality just cause that's where other emote menu settings were. 

Since closing menu is now optional, I thought it would be a good idea to at least auto close the menu by default when a chat message is sent. Personally I really don't see any reason why a user would want the emote menu to stay open after sending the message, so I think it's a good default behavior. 

The way I went about it kinda felt a bit hacky though, so would love feedback on better ways to know when a chat message has been sent. I'm new to Vue and the codebase so prob missing something.

AnatoleAM:
Closing as stale, too many conflicts.

PR #545, opened 2023-04-27
feat: add username highlight

### LINKED ISSUE #473: [Request] Hope can be added Custom Highlights user-name function
Hello,
Want that in the Custom Highlights function, can be added a check option for searching user-name

Judging from the current existing functions, this requirement should be easily fulfilled, but I have not learned the relevant grammar, so I am not sure if what I wrote is correct. 
And my computer has been problems in the environment installation, and there is no way to start "yarn start".
I Hope to add this feature in the future

in src/composable/chat/useChatHighlights.ts
image

in src/site/global/settings/SettingsConfigHighlights.vue
image

samnap11:
This PR is created to add a custom highlight feature for usernames.

### What this PR does
- Adds a new field of `highlightType` to highlight
- Adds a new field (column) to the highlight setting menu
- Adds a dropdown to let users pick which highlight type they want. However, it's **unstyled**.
- Adds a very simple dropdown component to support it
- Increases the width of the settings area. However, the dropdown (the select element) is still a bit concatenated because I don't want to increase the width too much.

The idea of using dropdown is from BTTV. If you have a better alternative, please do tell me.

If you're still unsatisfied with some of these, I'd be happy to discuss and improve these changes!

Final outlook:
Screenshot 2023-04-27 at 15 23 27

Closes #473

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -4,6 +4,7 @@
 
 -   Added a tooltip to show the full message when hovering over replies in chat
 -   Fixed tooltips of nametag paints appearing even if they are disabled
+-   Added an option to highlight chat messages from specific usernames
 
 ### Version 3.0.6.1000
 
--- a/src/composable/chat/useChatHighlights.ts
+++ b/src/composable/chat/useChatHighlights.ts
@@ -20,6 +20,7 @@ export interface HighlightDef {
 	regexp?: boolean;
 	readonly cachedRegExp?: RegExp;
 
+	highlightType?: "username" | "message";
 	color: string;
 	label: string;
 	caseSensitive?: boolean;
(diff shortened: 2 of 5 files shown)


lumynou5:
How about putting username highlight options seperately rather than in the same place? Because if people wanna highlight by username, they usually want to highlight messages from the exact users instead of all users with a username matching the pattern.

ayyybubu:
0iVWO0K
Would be nice to match the current dropdown style

samnap11:
@bennybanano That seems to be a really nice feature, but I think priority won't be included in this PR. I want to focus on creating a basic username highlight. Maybe after 7TV has implemented all kinds of highlights, then we can add the priority feature.

lumynou5:
Your change to `checkMatch` results many duplicates, I think if you want to make this feature by adding a field for highlight type, you may try like instead:
```ts
const str = h.highlightType == 'username' ? h.author.username : h.body;
// ...
```

TroyKomodo:
Features against this repo wont be implemented, we are going to rewrite the extension on this [repo](https://github.com/SevenTV/SevenTV).

PR #547, opened 2023-04-28
Added emojis as icons for the emoji groups

ZKLaurin:
Changed white Smileys to Emojis representing the group.
image

### DIFF
--- a/src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
+++ b/src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
@@ -28,11 +28,11 @@
 						@click="select(es.id, $refs['es-' + es.id] as InstanceType<typeof EmoteMenuSet>[])"
 					>
 						<div class="set-sidebar-icon">
-							<div v-if="es.provider === 'EMOJI' && emojiCategories.indexOf(es.name) > -1" class="emoji-group">
-								<SingleEmoji
-									:id="emojiGroupIcons[emojiCategories.indexOf(es.name)]"
-									:alt="es.name"
-								/>
+							<div
+								v-if="es.provider === 'EMOJI' && emojiCategories.indexOf(es.name) > -1"
+								class="emoji-group"
+							>
+								<SingleEmoji :id="emojiGroupIcons[emojiCategories.indexOf(es.name)]" :alt="es.name" />
 							</div>
 							<img v-else-if="es.owner && es.owner.avatar_url" :src="es.owner.avatar_url" />
 							<Logo v-else class="logo" :provider="es.provider" />
--- a/src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
+++ b/src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
@@ -98,7 +98,7 @@ const favorites = useConfig<Set<string>>("ui.emote_menu.favorites");
 // const shouldShowUsage = useConfig<boolean>("ui.emote_menu.most_used");
 
 // Emote icons for groups
-const emojiGroupIcons = ["1f603", "1f44b", "1f43b", "1f354", "26bd", "1f698", "1f4a1", "1f523", "1f6a9"];
+const emojiGroupIcons = ["1f603", "1f44b", "1f343", "1f354", "26bd", "1f698", "1f4a1", "1f523", "1f6a9"];
 const emojiCategories = [
 	"Smileys & Emotion",
 	"People & Body",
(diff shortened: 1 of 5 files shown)


src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
```diff
@@ -90,6 +98,20 @@ const favorites = useConfig<Set<string>>("ui.emote_menu.favorites");
 // const usage = useConfig<Map<string, number>>("ui.emote_menu.usage");
 // const shouldShowUsage = useConfig<boolean>("ui.emote_menu.most_used");
 
+//Emote icons for groups
```
AnatoleAM, line 101: nit: add whitespace at comment start

```suggestion
// Emote icons for groups
```

src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
```diff
@@ -29,6 +29,13 @@
 					>
 						<div class="set-sidebar-icon">
 							<img v-if="es.owner && es.owner.avatar_url" :src="es.owner.avatar_url" />
+							<div v-else-if="es.provider === 'EMOJI'" class="emoji-group">
+								<SingleEmoji
+									:id="emojiGroupIcons[emojiCategories.indexOf(es.name)]"
+									:alt="es.name"
+									:style="{ width: `100%`, height: `100%` }"
+								/>
+							</div>
```
AnatoleAM, line 38: nit: it makes more sense to invert this condition; while it's unlikely `owner` would ever be defined on an emoji set, if it ever was then this means the special emoji case wouldn't render.

```suggestion
							<div v-if="es.provider === 'EMOJI'" class="emoji-group">
								<SingleEmoji
									:id="emojiGroupIcons[emojiCategories.indexOf(es.name)]"
									:alt="es.name"
									:style="{ width: `100%`, height: `100%` }"
								/>
							</div>
							<img v-else-if="es.owner && es.owner.avatar_url" :src="es.owner.avatar_url" />
```

src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
```diff
@@ -29,6 +29,13 @@
 					>
 						<div class="set-sidebar-icon">
 							<img v-if="es.owner && es.owner.avatar_url" :src="es.owner.avatar_url" />
+							<div v-else-if="es.provider === 'EMOJI'" class="emoji-group">
+								<SingleEmoji
+									:id="emojiGroupIcons[emojiCategories.indexOf(es.name)]"
+									:alt="es.name"
+									:style="{ width: `100%`, height: `100%` }"
```
AnatoleAM, line 36: The inline style may not be necessary as you already define a class in the outer div. Target this element within it

ZKLaurin: I removed this styling in commit 1dda7d0

src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
ZKLaurin: Changed the order in 1dda7d0 .
However, I don't know if this is the right thing to do, because you have to specify a link in the owner that you probably want to display. It should be fine either way for this special scenario and I trust you have more experience for this stuff.

src/site/twitch.tv/modules/emote-menu/EmoteMenuTab.vue
ZKLaurin: Added whitespace in 1dda7d0

AnatoleAM (commented):
Nice, maybe though the categories could be sorted according to standard?

Which is, I believe:
- Smileys & Emoticon
- People & Body
- Animals & Nature
- Food & Drink
- Activities
- Travels & Places
- Objects
- Symbols
- Flags

ZKLaurin:
I added a condition in the sorting function to sort after the `emojiCategories` Array.
This was the only place i could really find to sort this, since the EmoteSet is a Record Type and is always sorted alphabetically.
Hopefully this approach is still fine with you :)

ZKLaurin:
I messed up the rebase, but nothing changed except the changelog...

src/site/twitch.tv/modules/hidden-elements/hiddenElements.ts
AnatoleAM, line 1: Why is this here?

ZKLaurin:
@AnatoleAM Sorry for the previous push, i reset it and merged it with the master branch.

PR #552, opened 2023-04-29
Added "focus on chatter" feature

DanielFirpo:
https://imgur.com/a/66m5PSQ

Added the ability to double click on a username in chat to dim all others, so you can easily see messages from that user. Also added setting to disable/enable.

I think it's planned to add a better user card that shows all of a user's messages, at which point this feature could be argued to be redundant, but I think it might still have a place as a more convenient way to quickly focus on a chatter for a few seconds to see if they respond to something, without needing to have the user card open on your screen. Idk, what do u think?

### DIFF
--- a/src/site/twitch.tv/modules/chat/ChatList.vue
+++ b/src/site/twitch.tv/modules/chat/ChatList.vue
@@ -403,7 +403,6 @@ function handleClick(e: MouseEvent) {
 		if (clickedElement.parentNode) {
 			if (clickedElement.parentNode.parentNode) {
 				const parentOfParent = clickedElement.parentNode.parentNode as Element;
-				console.log(parentOfParent);
 				if (parentOfParent.classList.contains("seventv-chat-user-username")) {
 					return;
 				}
--- a/src/site/twitch.tv/modules/chat/ChatList.vue
+++ b/src/site/twitch.tv/modules/chat/ChatList.vue
@@ -7,7 +7,12 @@
 					v-bind="{ msg }"
 				>
 					<component :is="msg.instance" v-bind="msg.componentProps" :msg="msg">
-						<UserMessage :msg="msg" :emotes="emotes.active" :chatters="messages.chattersByUsername" />
+						<UserMessage
+							:msg="msg"
+							:emotes="emotes.active"
+							:chatters="messages.chattersByUsername"
+							:focusedChatters="focusedChatters.ref"
+						/>
 					</component>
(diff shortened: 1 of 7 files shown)


src/site/twitch.tv/modules/chat/ChatList.vue
```diff
@@ -382,6 +386,30 @@ function onMessageIdUpdate(msg: Twitch.IDUpdateMessage) {
...
+			if (clickedElement.parentNode.parentNode) {
+				const parentOfParent = clickedElement.parentNode.parentNode as Element;
+				if (parentOfParent.classList.contains("seventv-chat-user-username")) {
+					return;
+				}
+			}
+		}
+	}
+	focusedChatters.value = [];
+}
```
AnatoleAM, line 411: Why are you doing DOM operations on our own elements?

src/site/twitch.tv/modules/chat/components/message/UserMessage.vue
```diff
@@ -251,6 +255,38 @@ watchEffect(() => {
...
+			}, 300);
+		} else if (ev.detail >= 2) {
+			// Double click
+			clearTimeout(clickTimer);
+			onUsernameDoubleClick();
+		}
+	} else {
+		openViewerCard(ev, msg.value.author!.username, msg.value.id);
+	}
+}
```
AnatoleAM, line 289: Such logic should be built into a [directive](https://vuejs.org/guide/reusability/custom-directives.html).

src/site/twitch.tv/modules/chat/components/message/parts/Mention.vue
```diff
@@ -25,12 +26,36 @@ const props = defineProps<{
...
+				tools.openViewerCard(ev, props.token.content.recipient, props.msg.id, target);
+			}, 300);
+		} else if (ev.detail >= 2) {
+			// Double click
+			clearTimeout(clickTimer);
+			onUsernameDoubleClick();
+		}
+	} else {
+		tools.openViewerCard(ev, props.token.content.recipient, props.msg.id);
+	}
```
AnatoleAM, line 58: Duplicate code, why?

AnatoleAM (changes requested):
This is unfortunately quite low quality, I would advise learning a little bit about Vue (the framework used by the Extension) first, then this could be integrated better now that we have a proper user card, for example a button inside the card to toggle this.

PR #555, opened 2023-04-30
[Bug Fix] React button overlap in hidden elements feature

ZemyCode:
## Bug Fix 

Small bug fix if hiding "Unfollow" button without hiding "React" button.

This PR also adds another selector for removing the live notification button with the goal of removing the left padding to make the UI look cleaner after removal. However, the original selector is still added as a fallback option since the new selector relies on a chain of multiple children.

### Screenshots
Normal: 
image

#### Before Fix:
With just unfollow hidden:
image

With unfollow and live notification hidden:
image

With just live notification hidden:
image

#### After Fix:
With just unfollow hidden:
image

With unfollow and live notification hidden:
image

With just live notification hidden: 
image

### DIFF
--- a/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
+++ b/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
@@ -196,6 +196,10 @@ export const config = [
 	button[data-test-selector="unfollow-button"] {
 		display: none !important;
 	}
+	// Prevents "React" button from overlapping the notification / subscribe button
+	div[data-target="channel-header-right"] > div:first-child {
+		margin-right: -1rem !important;
+	}
 }
 
 .seventv-hide-live-notification-button {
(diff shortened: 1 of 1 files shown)


ZemyCode:
@AnatoleAM I discovered this bug right after testing in the nightly version... should be a quick fix though.

src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
```diff
@@ -196,12 +196,25 @@ export const config = [
...
 	}
+	// Remove entire div containing live noti button to get rid of left padding
+	div[data-target="channel-header-right"]
+		> *:nth-child(2)
+		> *:nth-child(1)
+		> *:nth-child(2)
+		> *:nth-child(1)
+		> *:nth-child(2) {
+		display: none !important;
+	}
```
AnatoleAM, line 217: This is likely to break, at this point it may be preferable to hook the react component itself rather than use purely CSS.

PR #558, opened 2023-04-30
feat/preferred-emoji-skin-type: Added emoji skin type preference

### LINKED ISSUE #549: [REQUEST] Emoji-variants option

### Feature Description

Currently, when enabling this setting:
image

Multiple variations of the same emoji appear:

image

This creates unneeded clutter, at least in my eyes. I see the point of having the option for color variants etc. but for regular usage, yellow/default works fine enough, and this is more of an annoyance than it is helpful to me.

I can imagine a toggle to not show variants, with the default keeping the variants shown for users that require this.

berghall:
Closes #549


AnatoleAM (commented):
This is not the correct way to implement skin tone filtering. I believe we may be missing the necessary data however; this is likely something that would have to change within the emoji generation script.

PR #559, opened 2023-05-01
feat/hide-more-elements

MisterXYCB:
Added the option to hide the following three things:

- Squad-Stream Button
- Uptime
- Viewer Count


src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
```diff
@@ -216,4 +234,23 @@ export const config = [
...
+		display: none !important;
+	}
+	svg[type="color-text-accessible-red"] {
+		display: none !important;
+	}
+}
+.seventv-hide-squad-stream-button {
+	a[href$="squad"] {
+		display: none !important;
+	}
```
AnatoleAM, line 254: Most of these selectors are too broad and would result in content being hidden where it shouldn't be

AnatoleAM (changes requested):

AnatoleAM:
Please open a new PR if feedback is addressed.

PR #561, opened 2023-05-01
pause chat while using mod slider

chipsTM:
Adds option to pause the chat scroll while using the mod slider


src/site/twitch.tv/modules/chat/components/mod/ModSlider.vue
```diff
@@ -95,6 +100,8 @@ const handleRelease = (e: PointerEvent): void => {
 
 	data.calculate(0);
 	(e.target as HTMLElement).releasePointerCapture(e.pointerId);
+
+	if (shouldPauseChat.value) scroller.unpause();
```
AnatoleAM, line 104: Check if the pause was triggered by the mod slider

AnatoleAM:
The issue with this change is that it risks creating conflicts in the future. Pausing chat must be extended, with something such as a reason field in order to avoid this.

Closing because this has been up for a long time but you are free to re-integrate this alongside a better pause method.

Issue #484, opened 2023-04-09
[feature] pause the stream/vod by clicking on player


equivalentx:
why can i unpause by clicking the player but can't pause? i could before i switched from bttv. annoying.
![image](https://user-images.githubusercontent.com/127750474/230775149-e090322c-93ce-412a-8dfe-bf5b2c9716a9.png)

TheRealGI:
To add more context. In BTTV there is a setting for this feature.  Open the BetterTTV Setting < Channel Settings  and scroll to the bottom.

PR #562, opened 2023-05-02
feat/pause video on click

ZemyCode:
## Feature

Reference: https://github.com/SevenTV/Extension/issues/484

This PR adds a "VideoPlayerModule" and with this module the feature to pause the video player on click is added.
Event listener is added on page load or when the setting is enabled, and the event listener is removed when setting is disabled.
Setting is false by default.

image

Added a sub category to "Channel" called "Video Player" for this setting. Would like feedback on if I should move to a different category or change the name at all. Possibly "General" instead of "Channel"?

### DIFF
--- a/src/site/twitch.tv/modules/video-player/VideoPlayerModule.vue
+++ b/src/site/twitch.tv/modules/video-player/VideoPlayerModule.vue
@@ -16,19 +16,19 @@ const videoPlayerClickHandler = document.querySelector('div[data-a-target="playe
 const video = document.querySelector('div[data-a-target="video-player"] video') as HTMLMediaElement;
 
 function pauseOnClick(): void {
-    if (video && !video.paused) {
+	if (video && !video.paused) {
 		video.pause();
 	}
 }
 
 watchEffect(() => {
-    if (videoPlayerClickHandler) {
-        if (pauseOnClickEnabled.value) {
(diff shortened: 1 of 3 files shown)


src/site/twitch.tv/modules/video-player/VideoPlayerModule.vue
```diff
@@ -0,0 +1,49 @@
...
+import { watchEffect } from "vue";
+
+const { markAsReady } = declareModule("video-player", {
+	name: "Video Player",
+	depends_on: [],
+});
+
+const pauseOnClickEnabled = useConfig<boolean>("channel.video_player.pause_on_click");
+
+const videoPlayerClickHandler = document.querySelector('div[data-a-target="player-overlay-click-handler"]');
```
AnatoleAM, line 15: Please don't use DOM operations. We have built an array of React utilities capable of hooking Twitch's components, you can see those under `common/ReactHooks.ts` and `common/Reflection.ts`.

In order to hook to hook the relevant component, a predicate such as this might work (untested, only visually found this scrolling through the react tree)
```ts
(n) => n.props && n.props.mediaPlayerInstance && n.props.videoContainer
```

AnatoleAM (changes requested):

AnatoleAM:
Superseded by #727

PR #582, opened 2023-05-13
Add link to emote in emote card

saliven:
Add an icon with a link to open the emote in a new tab directly from the emote card.


AnatoleAM (approved):
The emote card pends a general redesign but this works for now :slightly_smiling_face:

PR #583, opened 2023-05-14
feat: Enable AVIFs on Firefox >= v113 by default

bl4d3tv:
Firefox [has supported AVIF images for a while](https://www.mozilla.org/en-US/firefox/93.0/releasenotes/), but it didn't support animated sequences. As of Firefox 113, [animated sequences are enabled by default](https://www.mozilla.org/en-US/firefox/113.0/releasenotes/).

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,7 +2,7 @@
 
 **The changes listed here are not assigned to an official release**.
 
-- No unreleased changes yet.
+-   Enable AVIF images on Firefox >= 113 by default
 
 ### Version 3.0.7.1000
 
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,7 +2,7 @@
 
 **The changes listed here are not assigned to an official release**.
 
--   Enable AVIF images on Firefox >= 113 by default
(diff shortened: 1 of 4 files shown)


AnatoleAM:
Note: the fact that Firefox supported AVIF static images, but did not actually support the full AVIF spec, is exactly the reason why we had to make this very specific check in the first place, rather than using a standard detection method such as `<picture>`.

Essentially, the browser was falsely claiming to support AVIF. It would've rendered animated images as static. Quite the PITA.

AnatoleAM (approved):

AnatoleAM (commented):
Can you bump `dev_version` to `2.0` in package.json please :slightly_smiling_face:

bl4d3tv:
> Note: the fact that Firefox supported AVIF static images, but did not actually support the full AVIF spec, is exactly the reason why we had to make this very specific check in the first place, rather than using a standard detection method such as `<picture>`.
> 
> Essentially, the browser was falsely claiming to support AVIF. It would've rendered animated images as static. Quite the PITA.

Yeah, I found it quite strange when they released AVIF support but didn't support animated sequences. Mozilla is kind of known to make weird decisions...

> Can you bump `dev_version` to `2.0` in package.json please 🙂

Done!

PR #611, opened 2023-05-17
Fixed slow-mode timer on twitch

### LINKED ISSUE #475: [Bug] Slow-Mode Timer on Twitch
Browser: Opera GX
7TV Version: 3.0.3

The 7TV version simply hides the slow mode timer.
So you have no chance to see when you are allowed to write again. 

Neither the number nor the animated running bar is displayed.

Batyodie:
Fixed #475.

There was a problem with the watcher in the component. src/site/twitch.tv/modules/chat/ChatList.vue
```typescript
// Keep track of props
// The message handler is hooked to render messages and prevent
// the native twitch renderer from rendering them
watch(
	messageHandler,
	(handler, old) => {
		if (handler !== old && old) {
			unsetPropertyHook(old, "handleMessage");
		} else if (handler) {
			defineFunctionHook(handler, "handleMessage", function (old, msg: Twitch.AnyMessage) {
				const ok = onMessage(msg);
				// this same problems
				if (ok) return old?.call(this, msg); // message was rendered by the extension
				// message was not rendered by the extension
				unhandled.set(msg.id, msg);
				return old?.call(this, msg);
			});
		}
	},
	{ immediate: true },
);
```
In a standard situation, we render messages in a chat using extensions, apparently at this moment the context is lost and the functionality with the work of the slow mod begins to break. The data before the display component simply does not come.

I tried explicitly returning `return old?.call(this, msg);` instead of `return ""` And this solved the problem of displaying the slow mode text.

I looked at the work of the chat and did not find any failures in the messages.

As it was before:

[screencast-newtab-2023.05.17-19_50_33.webm](https://github.com/SevenTV/Extension/assets/61266300/281b3122-47b8-4829-9d37-4cec4be2640b)

How is it now:

[screencast-newtab-2023.05.17-19_53_12.webm](https://github.com/SevenTV/Extension/assets/61266300/84979019-d1ca-41fd-9a59-34636d1e8a09)


Batyodie:
@AnatoleAM Сan you take a look here too please

AnatoleAM:
If a more suitale solution is found open a new pull request.

lucas-teks:
Hi, thanks for the fix but this issue is still appearing in latest 7TV version both nightly and stable releases, both on latest version of Chrome and Firefox with default configuration.
Slow mode bar doesn't show up, and entering a message before the delay will "fake" send it, the server returning in the chatbox I'm not allowed to chat yet.

PR #627, opened 2023-05-20
Highlight messages from specific user

vanedei:
Added ability to highlight messages from specific user in a chat. Username should be specified in `Pattern` field. When `Username` checkbox is checked, `Pattern` field value will be changed to lowercase, and checkboxes `RegExp` and `Case Sensitive` would be unchecked and disabled.

image


lumynou5:
Isn't this a duplicate of #545?

PR #634, opened 2023-05-24
fix: correct link to user profile in user-card

Satont:
If we located in `Stream Manager` and opening user card, with origin we going to `dashboard.twitch.tv/{username}`, which is wrong link.

So, i hardcoded twitch url to correctly handle this.

I'm not sure does this component used by youtube, if yes tell me i'll add checks for site address in this function.

### DIFF
--- a/src/app/chat/UserCard.vue
+++ b/src/app/chat/UserCard.vue
@@ -383,7 +383,7 @@ function openNativeCard(ev: MouseEvent): void {
 }
 
 function getProfileURL(): string {
-	return window.location.origin + "/" + props.target.username;
+	return "https://www.twitch.tv/" + props.target.username;
 }
 
 function formatDateToString(date?: string): string {
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,7 +2,7 @@
 
 **The changes listed here are not assigned to an official release**.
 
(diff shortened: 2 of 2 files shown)


src/app/chat/UserCard.vue
```diff
@@ -383,7 +383,7 @@ function openNativeCard(ev: MouseEvent): void {
 }
 
 function getProfileURL(): string {
-	return window.location.origin + "/" + props.target.username;
+	return "https://www.twitch.tv/" + props.target.username;
```
AnatoleAM, line 386: This shouldn't be hardcoded, as there is no guarantee the current origin is `twitch.tv`. Use location data instead.

AnatoleAM (changes requested):

src/app/chat/UserCard.vue
crashmax-dev: > This shouldn't be hardcoded, as there is no guarantee the current origin is `twitch.tv`. Use location data instead.

I propose this solution to the problem

```ts
function getLocationOrigin(): string {
  if (location.hostname.includes("dashboard")) {
    return "https://www.twitch.tv/";
  }

  return location.origin;
}
```

AnatoleAM:
Closing as stale. If fixes are made open a new pull request

PR #608, opened 2023-05-16
add media-player module

chipsTM:
This PR aims to add the foundational work of working with the media player by combining several requested features into one PR.

Player Settings Section
image

Stream info tooltip
image 

Showcasing Twitch's auto speed up 
image

Successful merge of this PR will close #485 and #562

PR #727, opened 2023-06-26
Video Stats & Player Click Events

chipsTM:
This is a refactor of #608 

Player Settings Section
image

Video Stats tooltip (clicking will open native video stats)
image


AnatoleAM (commented):
There was quite a lot wrong... though I like the idea of this feature so I've went ahead and cleaned it up to expedite a merge. Hard to list exactly everything that conflicted with our quality standards so feel free to review the changes made.

src/site/twitch.tv/modules/player/PlayerController.vue
```diff
@@ -3,22 +3,25 @@
...
+	mediaPlayer?: Twitch.MediaPlayerInstance;
 }>();
 
 const shouldHideContentWarning = useConfig<boolean>("player.skip_content_restriction");
+const actionOnClick = useConfig<number>("player.action_onclick");
+
 const contentWarning = ref<ReactComponentHook<Twitch.VideoPlayerContentRestriction> | null>(null);
+const playerOverlay = ref<HTMLElement | null>(null);
 
 props.inst.component.props.containerRef.classList.add("seventv-player");
```
Melonify, line 26: Would be nice here to do a check first and log to the console, if the geometry changes here this will fail the whole component. Could be done in a later PR though.

src/site/twitch.tv/modules/player/PlayerModule.vue
```diff
@@ -1,32 +1,67 @@
...
 				return cur;
 			},
 		},
 	},
 );
+
+// hook for the media player instance advanced controls
+const playerAdvancedOptionsComponent = useComponentHook<Twitch.MediaPlayerAdvancedControls>({
+	parentSelector: ".persistent-player",
+	predicate: (el) => el.props && el.setStatsOverlay,
```
Melonify, line 52: Would be nice if these predicates had another prop for specificity, not strictly necessary though.

src/site/twitch.tv/modules/player/PlayerModule.vue
```diff
@@ -1,32 +1,67 @@
...
+});
+
+// hook to render video stats in the channel page view
+const streamInfo = useComponentHook<Twitch.StreamInfo>(
+	{
+		parentSelector: "#live-channel-stream-information",
+		predicate: (el) => el.props && el.props.liveSince,
```
Melonify, line 59: Would be nice if these predicates had another prop for specificity, not strictly necessary though.

Melonify (approved):
Other than the comments the rest LTGM :+1:

AnatoleAM (approved):
Thank you for the initiative on this feature :slightly_smiling_face: reply with your username on 7tv.app if you'd like to receive the contributor badge

chipsTM:
🥳 Glad to contribute! I have the same name on 7tv: chipsTM

dpthegrey:
I enabled the option but still cannot see the stats

image

chipsTM:
Does it work after a hard refresh (ctrl + F5) ?

dpthegrey:
> Does it work after a hard refresh (ctrl + F5) ?

yes did hard refresh (cmd + shift + R) in mac, still shows same as above image. no stats.

chipsTM:
> 7TV version: 3.0.7.2000 nightly

This version isn't really applicable as this feature is only available on hosted mode (at the moment)
image

I can't reproduce using same browser version (64-bit instead of arm64). I suspect there's something else preventing it from showing

dpthegrey:
@chipsTM I opened 7TV dashboard on twitch page and went to Home to check the version there, turns out I am also using the same version v3.0.13.3000 as you. 

I also tried in Firefox Developer Edition (version 115.0b9) and still no stats there.
I suspect this could be an arm64 issue then. Maybe someone using arm64 can confirm?

image

dpthegrey:
@chipsTM So I just checked on other device with x86_64 intel macbook and I can see stats 

image

chipsTM:
@dpthegrey I think at this point it would be best to create a bug issue specific for this so that a fix can be explored

PR #734, opened 2023-06-27
Refactor 7tv Settings Button and Hide Top Nav Buttons

chipsTM:
As the title implies, adds an option to hide the Turbo Button in the top nav menu.

Additionally, switch the bits button selector to make the spacing cleaner when hidden

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -15,8 +15,9 @@
 -   Fixed an issue which squished tooltips when hovering an emote on the far right side of chat
 -   Fixed an issue which hid historical messages on Kick
 -   Increased the default value for Message Batching from 150 to 250
--   Added an option to hide turbo button (when logged in)
--   Fixed css selector for hiding bits button in top nav menu
+-   Added component hook for top nav bar to toggle bits, prime, whispers, notifcations, and turbo buttons
+    -   update 7tv settings button to use hook
+    -   removed relevant css selectors for the same
 
 ### Version 3.0.13.1000
 
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -16,8 +16,8 @@
 -   Fixed an issue which hid historical messages on Kick
 -   Increased the default value for Message Batching from 150 to 250
 -   Added component hook for top nav bar to toggle bits, prime, whispers, notifcations, and turbo buttons
-    -   update 7tv settings button to use hook
-    -   removed relevant css selectors for the same
+    -   Updated 7TV settings button to use hook
+    -   Removed relevant css selectors in hidden elements module for the same
 
 ### Version 3.0.13.1000
 
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -15,6 +15,8 @@
 -   Fixed an issue which squished tooltips when hovering an emote on the far right side of chat
 -   Fixed an issue which hid historical messages on Kick
 -   Increased the default value for Message Batching from 150 to 250
+-   Added an option to hide turbo button (when logged in)
+-   Fixed css selector for hiding bits button in top nav menu
 
 ### Version 3.0.13.1000
 
--- a/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
+++ b/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
@@ -188,7 +188,7 @@ export const config = [
 
 .seventv-hide-bits-buttons {
 	button[data-a-target="bits-button"] {
-	// button[data-a-target="top-nav-get-bits-button"] {
+		// button[data-a-target="top-nav-get-bits-button"] {
 		display: none !important;
 	}
 }
--- a/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
+++ b/src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
@@ -111,6 +111,18 @@ export const config = [
 		hint: "If checked, the 'Get Ad-Free' button on the top bar will be hidden",
 		defaultValue: false,
 	}),
+	declareConfig("layout.hide_site_notifications_button", "TOGGLE", {
+		path: ["Site Layout", "Twitch Features"],
+		label: "Hide Site Notifications Button",
+		hint: "If checked, the 'Site Notifications' button on the top bar will be hidden",
+		defaultValue: false,
+	}),
+	declareConfig("layout.hide_whispers_button", "TOGGLE", {
+		path: ["Site Layout", "Twitch Features"],
+		label: "Hide Whispers Button",
+		hint: "If checked, the 'Whispers' button on the top bar will be hidden",
+		defaultValue: false,
+	}),
 	// Side bar elements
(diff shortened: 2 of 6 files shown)


AnatoleAM (changes requested):

src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
```diff
@@ -169,7 +176,7 @@ export const config = [
 
 .seventv-hide-bits-buttons {
 	button[data-a-target="bits-button"],
-	button[data-a-target="top-nav-get-bits-button"] {
+	.top-nav__menu > div:last-child > div:nth-child(5) {
```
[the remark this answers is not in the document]
chipsTM: Ironically, hooking the top nav bar breaks the settings button (since settings button uses manual dom operations)

chipsTM:
This is functionally working, however I'm taking feedback on how to structure/organize the code since there is shared functionality between modules

chipsTM:
I'm running into an issue where the component mounts too early, but also at the same time I want it to mount if it can't find the hooked instance. Maybe can help suggest what should be done here as this is a little beyond my current knowledge

src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
```diff
@@ -141,6 +159,7 @@ export const config = [
 </script>
 
 <style lang="scss">
+/* stylelint-disable selector-class-pattern */
```
Melonify, line 162: We should not be abusing lint disables to try to get around formatting.

src/site/twitch.tv/modules/hidden-elements/HiddenElementsModule.vue
```diff
@@ -248,14 +267,12 @@ export const config = [
 }
 
 .seventv-hide-player-ext {
```
Melonify, line 269: Instead here, rather instead of disabling stylelint rules, just use the encapsulation formatting

```suggestion
.seventv-hide-player-ext {
    .video-player {
      .extension-taskbar, .extension-container,
	  .extensions-dock__layout, .extensions-notifications,
	  .extensions-video-overlay-size-container {
		display: none !important;
	  }
	}
}
```

src/site/twitch.tv/modules/settings/SettingsModule.vue
```diff
@@ -20,6 +21,17 @@ const { markAsReady } = declareModule("settings", {
 
 const ctx = useSettingsMenu();
 
+// hook for the top nav bar
+const topnav = useComponentHook<Twitch.TopNavBarComponent>(
+	{
+		predicate: (n) => n.props && n.props.isWhispersLoaded && n.props.languageCode,
```
Melonify, line 27: This predicate should have a class selector

src/site/twitch.tv/modules/settings/SettingsMenuButton.vue
```diff
@@ -11,37 +11,103 @@
...
+	renderWhispers: hideWhispersButton,
+});
+
+// force update top nav when user changes settings
+watch(menuElements, (renderElements) => {
+	rerenderTopNav(renderElements);
+});
+
+// TODO: define proper types here
+function rerenderTopNav(els) {
```
Melonify, line 51: I don't really understand this at all, why are we trying to redefine the hook multiple times after each settings change?

CHANGELOG-nightly.md
```diff
@@ -15,6 +15,9 @@
 -   Fixed an issue which squished tooltips when hovering an emote on the far right side of chat
 -   Fixed an issue which hid historical messages on Kick
 -   Increased the default value for Message Batching from 150 to 250
+-   Added component hook for top nav bar to toggle bits, prime, whispers, notifcations, and turbo buttons
+    -   update 7tv settings button to use hook
```
Melonify, line 19: Change-log entries should be capitalized properly.

src/site/twitch.tv/modules/settings/SettingsMenuButton.vue
```diff
@@ -11,37 +11,103 @@
...
+		navmenu.insertBefore(containerEl.value, navmenu.lastChild);
+	}
 }
+
+// we need to render the current state from the user settings
+// this should only run once after mount
+watch(
+	() => props.inst,
+	() => {
+		rerenderTopNav(menuElements);
```
Melonify, line 103: This is a bit non descriptive as a function, you should probably do the checks for existence here so its a bit more clear whats going on.

src/site/twitch.tv/modules/settings/SettingsMenuButton.vue
```diff
@@ -11,37 +11,103 @@
...
+		if (!navmenu) {
+			// Header if popout
+			navmenu = document.querySelector(".stream-chat-header")?.lastChild;
+			tooltip = "left";
+		}
+	}
+	if (!navmenu) return;
+	if (navmenu.contains(containerEl.value)) {
+		// we have to re-insert the 7tv button since re-rendering will change order
+		navmenu.parentElement?.querySelector<HTMLElement>("#seventv-settings-button")?.remove();
```
Melonify, line 91: This is very strange and doesn't make sense, as well as being very dangerous if it worked as you expected it to (your modifying a vue element directly).

Why do this at all? You're traversing upwards only to query the element you already have a reference to, `containerEl.value` and `el` both refer to the element being queried here.

src/site/twitch.tv/modules/settings/SettingsMenuButton.vue
```diff
@@ -11,37 +11,103 @@
...
+watch(
+	() => props.inst,
+	() => {
+		rerenderTopNav(menuElements);
+	},
+);
+
+// this re-inserts the button when we force update the top nav component
+watchEffect(() => {
+	insert7tvButton();
```
Melonify, line 109: This feels dangerous to me, I'm not sure this is the right approach, I'd rather you just watched `domNodes.root` with `watch(() => props.inst.domNodes, () => {})`

Melonify: If you're having issues with timing, it makes more sense to do this in the `update` hook of the component. I'm not sure watching the instance's root node will even work as you expect it to, as that would only trigger if the root itself did indeed change, however this can only occur if React reconciliation fails which would be rare.

`update` runs after React has flushed the VDOM, and it is safe to do your DOM querying there, just use `defineExpose` on this vue component, and use a `ref=` in `SettingsModule.vue` to call this function from outside in the module.

Melonify (changes requested):
The code quality here is slightly subpar, however I believe your current approach is correct, but needs refining.

src/site/twitch.tv/modules/settings/SettingsMenuButton.vue
Melonify: Shouldn't this function just call `forceUpdate()`?

src/site/twitch.tv/modules/settings/SettingsModule.vue
chipsTM: I believe the parent is the root, so I am not sure if specifying a selector would make a difference here

AnatoleAM:
Closing as stale

PR #736, opened 2023-06-28
Exclude self from mention & reply highlights

### LINKED ISSUE #654: [BUG] Mention highlight when tagging myself or replying to someone that replied to me

### This issue exists in the latest nightly version

- [X] I am using the latest nightly version

### What browsers are you seeing the problem on?

Opera

### Current Behavior

I get a mention highlight when replying to someone that replied to me or when I @ myself

### Expected Behavior

I expect to not get a mention highlight when replying to someone that replied to me or when I @ myself

### Steps To Reproduce

Have mention highlights turned on and @ yourself or reply to someone who replied to one of your messages

2547techno:
Fixes #654

Do not highlight messages as "reply" if message and parent message are from yourself (ie. sending a message in a thread starting from your own message)

Ignore mentions from yourself

~~This PR also allows the use of the `test` arg even when a regex patter is defined. If both a pattern and a test are defined, the result is AND'd. If only a test is defined, the result is the result of the test~~


src/composable/chat/useChatHighlights.ts
```diff
@@ -168,8 +169,13 @@ export function useChatHighlights(ctx: ChannelContext) {
...
 				: msg.body.toLowerCase().includes(h.pattern.toLowerCase());
-		} else if (typeof h.test === "function") {
-			ok = h.test(msg);
+		} else {
+			justTest = true;
+		}
+
+		if (typeof h.test === "function") {
+			const t = h.test(msg);
+			ok = justTest ? t : ok && t;
```
AnatoleAM, line 178: What is the purpose of this?

2547techno: When defining a `chatHighligh`, if `regexp`/`pattern` are defined, you can still define `test`, both must pass to return `true` for `checkMatch`.

Originally if both were defined, `test` would be ignored due to the `else if`

AnatoleAM: That doesn't seem useful, there is no reason I can think of where it would be desirable to have both a pattern and a test function. The latter is capable of doing the former anyway.

2547techno: Changes made

PR #739, opened 2023-06-29
Export & Import Settings

### LINKED ISSUE #694: [REQUEST] Ability to Import/Export user configuration

### Feature Description

Some way to Export/Backup configured settings to a File, Such that if/when we change Browser/Device we can directly import Settings/Configuration based on previous install. 

Also, saves time to reconfigure extension and potential config mismatch if we do manually.

This feature exists on [BTTV](https://betterttv.com/) and [FFZ](https://www.frankerfacez.com/), and there is a need for it on [7TV](https://7tv.app/) browser extension also.

2547techno:
Closes #694

Added new settings tab to import & export settings

Draft:
> Looking for some feedback on the initial approach. Would this be something that could be used to moved forward with?
> 
> Also wasn't sure about the general feeling of adding a new dependency.
> 
> Placeholder icon for Backup section is from Font Awesome Free so I assume that needs to be changed before any merge.
> 
> - [x] Serialize settings into JSON file
> - [x] Prompt browser to download JSON file
> - [x] Load JSON file
> - [x] Deserialize loaded JSON & type check
> - [x] Overwrite settings store


Melonify:
This is a good idea, however might not work in all cases, not sure a good way around that either. It wont be able to save some things, like the custom audio files (I think there may be some other strange objects used as well, as the settings system doesn't really make distinctions of types)

Also be weary as this may save MORE than you expect it to as well, some internals use settings as storage, you may be well off adding another defaulted parameter for if a setting should be attempted to be saved. 

Be sure to strip away any Vue reactivity before serializing as well if you don't already, you don't want to accidentally store the internal props Vue exposes in its proxies like `__markedRaw` and `_shouldSkip`.

2547techno:
### Changes made: Non-primitive constructor types now need to have a defined deserialization function for them to be allowed to be serialized/deserialized.

https://github.com/SevenTV/Extension/blob/b9dfedb569706849a877fa411777a68cf1f40906/src/composable/useSettings.ts#L15-L24

> It wont be able to save some things, like the custom audio files

I wasn't able to find the setting that stored an audio file or custom objects, but in the future if a setting like this were introduced, you would now have to explicitly define it's deserialization function.

> Also be weary as this may save MORE than you expect it to as well, some internals use settings as storage

Just to make sure, are you referring to these type of settings?

https://github.com/SevenTV/Extension/blob/b9dfedb569706849a877fa411777a68cf1f40906/src/site/twitch.tv/modules/emote-menu/EmoteMenuModule.vue#L103-L117

> Be sure to strip away any Vue reactivity before serializing as well if you don't already, you don't want to accidentally store the internal props Vue exposes in its proxies like __markedRaw and _shouldSkip.

The DB response doesn't include them anyways 👍 

@Melonify thanks for the feedback!

2547techno:
https://github.com/SevenTV/Extension/blob/fc23a8a52d9856777f7405e3377197dd8b14dbcd/src/site/twitch.tv/modules/emote-menu/EmoteMenuModule.vue#L114-L119

Now you can skip serialization for setting by adding an option when declaring the setting.

Only skips serialization if `serialize` is strictly `false`. Assume `undefined` as `true`

2547techno:
Page preview:
image

Screenshot includes dummy settings to show what will display when there are unserializable settings of both types, from either the forced `serialize: false` option or a setting with an undefined deserialization function, `URL` in this case
https://github.com/SevenTV/Extension/blob/e0b407e261665ed8fe67677fd59f07066401361a/src/site/twitch.tv/modules/player/PlayerModule.vue#L75-L87

Will mark as ready for review after removing dummy settings & updating changelog

2547techno:
forgot to add ss of error message 
image

package.json
```diff
@@ -61,6 +65,7 @@
 		"eslint": "^8.43.0",
 		"eslint-plugin-prettier": "^4.2.1",
 		"eslint-plugin-vue": "^9.15.0",
+		"file-saver": "^2.0.5",
```
AnatoleAM, line 68: You can simply use `URL.createObjectURL()`, create a link element and trigger a click on it without any extra dependency needed.

AnatoleAM: The rest seems fine, just clear out this unnecessary dep.

src/app/settings/SettingsViewBackup.vue
```diff
@@ -0,0 +1,146 @@
...
+<script setup lang="ts">
+import { ref } from "vue";
+import { useStore } from "@/store/main";
+import { log } from "@/common/Logger";
+import {
+	SerializedSettings,
+	deserializeSettings,
+	exportSettings as e,
+	getUnserializableSettings,
+	importSettings as i,
```
AnatoleAM, line 30: nit: what is the purpose of this alias? (readability decreased)

src/app/settings/SettingsViewBackup.vue
```diff
@@ -0,0 +1,146 @@
...
+</template>
+
+<script setup lang="ts">
+import { ref } from "vue";
+import { useStore } from "@/store/main";
+import { log } from "@/common/Logger";
+import {
+	SerializedSettings,
+	deserializeSettings,
+	exportSettings as e,
```
AnatoleAM, line 28: nit: what is the purpose of this alias? (readability decreased)

src/app/settings/SettingsViewBackup.vue
```diff
@@ -0,0 +1,146 @@
...
+			res(fileReader.result as string);
+		};
+		fileReader.onerror = (err) => rej(err);
+	});
+}
+</script>
+
+<style scoped lang="scss">
+main.seventv-settings-backup {
+	display: flex;
```
AnatoleAM, line 113: nit: CSS Grid is preferable over Flexbox. Please consider using `grid` instead.

src/composable/useSettings.ts
```diff
@@ -3,12 +3,44 @@ import { log } from "@/common/Logger";
...
+};
+// const { platform } = useStore();
+/*
+Uncaught (in promise) Error: [🍍]: "getActivePinia()" was called but there was no active Pinia. Did you forget to install pinia?
+	const pinia = createPinia()
+	app.use(pinia)
+This will fail in production.
+    at useStore (pinia.mjs:1699:19)
+    at useSettings.ts:14:15
+*/
```
AnatoleAM, line 33: nit: leftover dev comment?

AnatoleAM: Pinia cannot be used outside of the Vue context.

AnatoleAM (approved):
Looks good thank you for the contribution :rocket: 

As a thank you for adding a feature to the 7TV Extension, please message me your username if you'd like to receive the Contributor chat badge. :slightly_smiling_face:

2547techno:
@AnatoleAM Thanks! my username on twitch & 7tv is `2547techno`

also thanks @Melonify for catching that 👍

2547techno:
> Thanks! my username on twitch & 7tv is `2547techno`

@AnatoleAM Hey, sorry, not sure if you missed this. I haven't opened chatterino in a while so I didn't realize it was missing. Any chance you can have a look?

PR #742, opened 2023-06-30
Add Artist Attribution to Emote Card

2547techno:
image


AnatoleAM (approved):
This works for now, noting the plan is to redesign this card entirely (or, design it at all, since it's more of a messy data table as-is)

PR #746, opened 2023-07-02
fix: UserCard "hidden subscription status" message

shaanaliyev:
The issue occurs when the "Hide Subscription Status in Chat Viewer Card" option is enabled in Twitch settings.

It says "Tier 1 - Subscribed for 0 months" no matter how long or which tier you have subscribed.

Before: 
before

After: 
after

This gives more accurate information about the subscription status.


AnatoleAM (approved):
This is fine, thank you :slightly_smiling_face:

Issue #576, opened 2023-05-11
[REQUEST] Reset setting to its' default value


berghall:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This feature does not exist in the latest nightly version

- [X] I have checked the latest nightly version

### Feature Description

I want to reset a changed setting to its' default value
Should there also be a reset all settings button?

![Group_184](https://github.com/SevenTV/Extension/assets/20751720/b9f62abd-d7f7-4e2d-8316-37bad8aa0e57)

Originally requested by @ayyybubu 👽

PR #758, opened 2023-07-06
settings(feat): reset to default value

chipsTM:
adds #576 

image

taking suggestions on css formatting and handling custom components/ignores

### DIFF
--- a/src/app/settings/SettingsNode.vue
+++ b/src/app/settings/SettingsNode.vue
@@ -10,13 +10,11 @@
 		<div class="label">
 			<div class="title" :class="{ unseen }">
 				{{ te(node.label) ? t(node.label) : node.label }}
-				<div
+				<CloseIcon
 					v-if="!ignoreDefaults.includes(node.key) && currentSetting != node.defaultValue"
 					class="reset-default"
 					@click="resetSetting"
-				>
-					x
-				</div>
+				/>
 			</div>
 			<div v-if="node.hint" class="subtitle">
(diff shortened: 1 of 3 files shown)


chipsTM:
I was thinking that another property (`disableReset` ?) should be added to `SettingNode` interface so that a setting can be declared as non-resettable when one uses `declareConfig`. At least this way we can exclude custom types or types with non-primitive values?

Melonify:
The behavior of this isn't quite right. Ideally it should remove the setting from the stored table entirely, so that if a setting's default value changes for some reason it's as if the user had never changed the setting.

Allowing the post-reset state to be separate from the initial state adds to potential user confusion.

src/app/settings/SettingsNode.vue
```diff
@@ -50,6 +58,14 @@ function onHover(): void {
 	useTimeoutFn(() => emit("seen"), 500);
 }
 
+const currentSetting = useConfig<SevenTV.SettingType>(props.node.key);
+
+// set the currentSetting back to default to trigger UI change before removing from the settings db
+function resetSetting() {
+	currentSetting.value = props.node.defaultValue;
+	db.settings.delete(props.node.key);
```
AnatoleAM, line 66: Please handle errors with `catch()` on the promise

AnatoleAM (changes requested):

AnatoleAM (approved):
Thank you for the contribution :slightly_smiling_face:

PR #788, opened 2023-07-21
Fix spacing between emote providers and emote set when live search is enabled

berghall:
257c0ceb768ff616bd55404e1b781f41


AnatoleAM (approved):
Thank you for the contribution :slightly_smiling_face:

PR #796, opened 2023-07-25
feat: Add check for favorite emotes and display helpful message

Apricosma:
If user does not have any emotes favorited, there is no indication on how to favorite an emote.
This PR adds a small display on the favorites menu to suggest Alt-clicking an emote to favorite one.

I honestly have no idea what I am doing and have never used vue before so I'm sure there's some terrible bug

## Issues?
I'm noticing provider issues and can't tell if it's my own code. Certain emotes won't show up when changing between streams/refreshing but the same behavior occurs when running main locally, I think it's just being blocked by CORS. If this is just a localhost issue then hopefully everything is fine...?

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -3,6 +3,7 @@
 **The changes listed here are not assigned to an official release**.
 
 -   Added an option to select alternating background color for chat messages
+-   Added a tip to the favorite menu to help users favorite emotes if none are found
 -   Fixed an issue with tab auto-completion on Kick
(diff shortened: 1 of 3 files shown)


AnatoleAM:
>  I think it's just being blocked by CORS. If this is just a localhost issue then hopefully everything is fine..

By default the developer environment points to localhost, expecting backend services to be present on certain ports. If you don't have these services on your PC, you can edit `.env.dev` with the values `env.production`. This is also something that should be added to the documentation

AnatoleAM (approved):

PR #799, opened 2023-07-27
Add indicator to show the settings sidebar has more options below

Melonify:
image

Disappears once the user has scrolled down.


AnatoleAM (commented):
I'm thinking that clicking the arrow should scroll you down, no?

Melonify:
@AnatoleAM Yeah I considered that, but I'm not sure if having the arrow block the pointer is a good idea or not. It might be annoying for the user to have to scroll before they can click on a tab if the tab is just barely underneath the icon.

Melonify:
But that does make more sense as its common a UX feature.

Issue #364, opened 2023-03-23
[Request] horizontal flip for all emotes with a suffix character modifier


nepeee:
Will be nice to have an option to horizontally flip(mirror) the emote showed in chat with just an extra character like peepoHappyM or other suffix character. So if the user types peepoHappyM instead of the original https://7tv.app/emotes/60b052fe3cadd71dffa73929 (right facing) the extension will display it left facing in chat.

PR #800, opened 2023-07-27
Added feature where if a suffix is added to an emotes it will horizontally flip the emote

nirajacharya2:
related feature https://github.com/SevenTV/Extension/issues/364 

after creating this PR i noticed few problems:
lets say the mirror suffix is `_M`
* if an emote already exist that has `_M` at the end it mirrors it. it shouldn't do that
* if a user wants to upload an emote with `_M` letters at end 7tv shouldn't let them upload them or at least warn them and i think there might be more requests like this for examples :
    * invert colors
    * make it continuously rotate
    * turn it 90,180 ... degrees
or even something like that we have natively in twitch like `emote_BW` makes black and white, `emote_SG`(probably zerowidth emote dose this) puts sunglasses etc. for that 7tv should reserve some suffix which might be annoying

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -8,6 +8,7 @@
 -   Fixed an issue where mentioning yourself would highlight the message
 -   Fixed an issue where replying to a thread starting from your own message would highlight the message
 -   Fixed an issue where emotes would take a long time to load if external emote providers gave slow response times
+-   Added a feature to horizontal flip emotes with a suffix character modifier
 
 ### 3.0.14.1000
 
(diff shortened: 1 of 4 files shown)


AnatoleAM:
This is not gonna be the way we implement this. For one, with the modifier token attached to the emote, it means any third party client will now only see text, which is double the amount of text pollution from how other providers have done similar, and from seeing their rollout, those features are rarely used after the initial week.

This would have to be significantly refined and provided with much more thought, a specification for third party implementation, and most likely require API integration allowing channels to turn this on or off.

PR #813, opened 2023-08-06
Alternating background colour

### LINKED ISSUE #738: [REQUEST] Allow users to pick their alternating background color / alpha

### Feature Description

As a Twitch chat reader
I would like to be able to pick the color of alternating background color / alpha
So that I can have more contrast making it easier to read and to have more freedom of choice

Right now were hard coding this value to `hsla(0deg, 0%, 50%, 6%)`.
It would be nice to be able to let the user pick the hsla value they want.
If this is too much freedom for the user, perhaps allowing them to just change the alpha percentage?

In my personal use case, I'd like to be able to set the alpha something higher so I can see more contrast (something around 15%, for example.)

2547techno:
Closes #738

image

Disabled when "Alternating Background" settings is disabled

Default color: `#808080` which was converted from HSL `hsl(0deg, 0%, 50%)`

Opacity is fixed at 6%, same as the original background value

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,6 +2,7 @@
 
 **The changes listed here are not assigned to an official release**.
 
+-   Added an option to select alternating background color for chat messages
 -   Fixed an issue with tab auto-completion on Kick
 -   Fixed emote tile width in emote menu
 
--- a/index.html
+++ b/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en" data-seventv-app>
 	<head>
 		<meta charset="utf-8" />
--- a/index.html
+++ b/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en" data-seventv-app>
 	<head>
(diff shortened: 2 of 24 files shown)


2547techno:
I don't get any linting errors locally, not sure why the gh action lint is failing. Should [547bb5f](https://github.com/SevenTV/Extension/pull/813/commits/547bb5fae3af3adcb646d7f6d5aac98899c91978) be reverted or are the linting rules misconfigured?

AnatoleAM:
> I don't get any linting errors locally

Make sure you've updated your dependencies (`make deps`)

2547techno:
ah ok, yea that was it, thanks

AnatoleAM (approved):
Thank you for the contribution.

Please reply with your twitch username if you'd like to receive the Contributor chat badge :slightly_smiling_face:

2547techno:
twitch username: 2547techno thanks! On Aug 20, 2023, at 9:22 PM, Anatole ***@***.***> wrote:﻿
@AnatoleAM approved this pull request.

Thank you for the contribution.
Please reply with your twitch username if you'd like to receive the Contributor chat badge 🙂

Issue #187, opened 2021-11-26
The browser plugin not working in the embed twitch chat iframe [Bug] 


nepeee:
**Describe the bug**
The browser plugin not working in the embed twitch chat iframe.

**To Reproduce**
<iframe src="https://www.twitch.tv/embed/[channel]/chat?darkpopout&parent=[parentdomain]"></iframe>
where [channel] is the name of the channel [parentdomain] is the domain/path of the site where the embed is placed.
Both bttv and ffz works.

**Expected behavior**
The plugin is working in the twitch embed chat :)

**Version:**
All current, checked in chrome and firefix.

dillon-cullinan:
Common use case where this bug comes into play is the use of alternative players for Twitch which are commonly used to get around Twitch ads.

AnatoleAM:
This was implemented by #193 and released with version [2.1.2](https://github.com/SevenTV/SevenTV/releases/tag/2.1.2).

HukilauKohi:
I am using version 2.2.2 for Chrome and am not able to see/use the 7TV plugin when viewing an embed chat.

In the image below the left side is chat when viewed from the streamers channel, while the right is viewed from: 
`https://www.twitch.tv/embed/[channel]/chat?darkpopout&parent=twitch.tv`


![Chat Embed](https://user-images.githubusercontent.com/105951275/173679126-17b1a7ce-6b1d-4441-a047-e10dc0a80a0a.png)

Should this be flagged as a new issue/bug, or am I missing something here?

tlstommy:
was this ever solved? Im getting the same issue embedding into an iframe

nepeee:
> was this ever solved? Im getting the same issue embedding into an iframe

It was but then broke again :(

PR #919, opened 2023-10-18
fix: added iframe support 

tlstommy:
added "all_frames":true, to manifest to support iframes

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -2,7 +2,7 @@
 
 **The changes listed here are not assigned to an official release**.
 
--   No unreleased changes yet
+-   Reinstated animated avatars
 
 ### 3.0.16.1000
 
--- a/src/site/twitch.tv/modules/avatars/AvatarsModule.vue
+++ b/src/site/twitch.tv/modules/avatars/AvatarsModule.vue
@@ -98,7 +98,7 @@ async function findAvatarClass() {
 
 	const com = getVNodeFromDOM(ael);
 	if (!com || !com.return) return;
-	if ((com.return.type as { displayName?: string }).displayName !== "ScAvatar") return;
+	if ((com.return?.return?.type as { displayName?: string }).displayName !== "Avatar") return;
 
 	avatarClass.value = com.return.type as unknown as typeof avatarClass.value;
 }


Melonify (changes requested):
We don't actually want to run on embeds, this would cause massive problems for sites that embed a lot of clips/embeds, like LSF.

It may be possible to check if the page is specifically a chat embed, and only inject then. But these are pretty rare, and I'm not sure that its worth the effort.

Melonify:
Is there a specific use-case where running in embeds is desirable?

tlstommy:
I have a chrome extension I'm using/developing that will inject other channels chats as an iframe onto a stream, and those embedded iframes won't show chat unless that's enabled.

Is there another way to do this? Or enable it so it will only allow for twitch.tv embeds?

I'd love the help if you are able!

tlstommy:
I would only need it to embed the twitch chat pop out

tlstommy:
Ive got it to work (i think) with the content script as this `(function () { "use strict"; 

//Check if it's an iframe and is a popout
if (window !== window.top) {
    let currentIframeSrc = window.location.href;
    if (!currentIframeSrc.startsWith("https://www.twitch.tv/popout")) {
        return;
    }
}



const r = "seventv-app-broadcast-channel"; async function d() { const e = document.createElement("div"); e.id = "seventv-emoji-container", e.style.display = "none", e.style.position = "fixed", e.style.top = "-1px", e.style.left = "-1px"; const t = chrome.runtime.getURL("assets/emoji"), i = 11; for (let s = 0; s < i; s++) { const c = (await fetch(t + "/emojis" + s + ".svg")).text(), n = document.createElement("div"); n.id = "emojis" + s, n.innerHTML = await c, e.appendChild(n) } document.head.appendChild(e) } const u = () => { const e = document.createElement("script"); e.src = chrome.runtime.getURL("site.js"), e.id = "seventv-extension", e.type = "module", e.setAttribute("worker_url", chrome.runtime.getURL("worker.js")), e.setAttribute("extension_origin", chrome.runtime.getURL("")); { const t = document.createElement("link"); t.rel = "stylesheet", t.type = "text/css", t.href = chrome.runtime.getURL("assets/seventv.style.3.0.15.css"), t.setAttribute("charset", "utf-8"), t.setAttribute("content", "text/html"), t.setAttribute("http-equiv", "content-type"), t.id = "seventv-stylesheet", (document.head || document.documentElement).appendChild(t) } (document.head || document.documentElement).appendChild(e), setTimeout(() => { d() }, 1e3) }, a = new BroadcastChannel(r); u(), a.addEventListener("message", e => { switch (e.data.type) { case "seventv-create-permission-listener": { const { selector: t, id: i, origins: s, permissions: c } = e.data.data, n = document.querySelector(t); if (!n) return; n.addEventListener("switch", () => { chrome.runtime.sendMessage({ type: "permission-request", data: { id: i, origins: s, permissions: c } }, {}, o => { o && a.postMessage({ type: "seventv-permission-granted", data: { id: o.id, granted: o.granted } }) }) }); break } case "seventv-update-check": chrome.runtime.sendMessage({ type: "update-check" }, t => { a.postMessage({ type: "seventv-update-check-result", data: { status: t.status, version: t.version } }) }) } }), chrome.runtime.onMessage.addListener(e => { switch (e.type) { case "update-ready": { m(e.data.version ?? ""); break } case "settings-sync": { p(e.data.settings ?? []); break } } }); function m(e) { a.postMessage({ type: "seventv-update-ready", data: { version: e } }) } function p(e) { a.postMessage({ type: "seventv-settings-sync", data: { nodes: e } }) } })();
 ` however im getting some trouble generating it.

tlstommy:
Ok, I added in a check so that even though all_frames is true, it also will check if it is a Twitch popout iframe before injecting. It works locally for me, but please let me know if there is something I should change!

tlstommy:
This would only look for twitch chat embed too, not any twitch embed

jaaneh:
Working on a site personally, and no emotes are working. Something like this would be pretty great. Pretty sad to not have emote support on embeds.

@tlstommy My chat embed URL looks like this, and wouldn't currently be picked up by your checker. `https://www.twitch.tv/embed/<channel>/chat`

tlstommy:
> Working on a site personally, and no emotes are working. Something like this would be pretty great. Pretty sad to not have emote support on embeds.
> 
> @tlstommy My chat embed URL looks like this, and wouldn't currently be picked up by your checker. `https://www.twitch.tv/embed/<channel>/chat`
> 
> Related #187 #193

It should detect it now!

tlstommy:
updated to new version, i think i formatted the nightly changelog correctly.

tlstommy (commented):
I have added logic to check specifically for chat embeds.

tlstommy:
https://github.com/SevenTV/Extension/pull/964

PR #937, opened 2023-11-18
Fix delete message functionality

MatinAniss:
Fixed the functionality of the delete message function to use Apollo instead of IRC, this conforms with how Twitch natively handles deleting messages. This will also fix issues where users are failing to delete messages seen in the image below.

Issue:
image

### DIFF
--- a/src/composable/chat/useChatModeration.ts
+++ b/src/composable/chat/useChatModeration.ts
@@ -1,13 +1,11 @@
 import { twitchBanUserQuery, twitchUnbanUserQuery } from "@/assets/gql/tw.chat-bans.gql";
+import { twitchDeleteMessageQuery } from "@/assets/gql/tw.chat-delete.gql";
 import { twitchPinMessageQuery } from "@/assets/gql/tw.chat-pin.gql";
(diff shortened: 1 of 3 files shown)


src/composable/chat/useChatModeration.ts
```diff
@@ -87,7 +85,18 @@ export function useChatModeration(ctx: ChannelContext, victim: string) {
 	}
 
 	function deleteChatMessage(msgID: string) {
-		messages.sendMessage(`/delete ${msgID}`);
+		const apollo = useApollo();
```
Melonify, line 88: Looks good, just one thing the `useApollo` call should probably be moved outside of this function and into the parent scope, the same way the `useChatMessages` composable was previously preserved.

Melonify (changes requested):

Melonify:
Looks good, just the one requested change is the only issue.

Melonify (approved):

Melonify:
LGTM :+1:

AnatoleAM (approved):

PR #938, opened 2023-11-18
Fix popup menu rendering improperly

MatinAniss:
This will fix the popup menu being rendered improperly on chrome browsers.

Issue:
image


MatinAniss:
oop, didn't see there was an existing PR that fixes this better. Don't know what this hasn't been fixed yet.

Issue #916, opened 2023-10-18
[REQUEST] Add "all_frames":true under "content_scripts" so that 7tv will work within an iframe.


tlstommy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This feature does not exist in the latest nightly version

- [X] I have checked the latest nightly version

### Feature Description

In the manifest, replace ` "content_scripts": [ {
      "js": [ "content.js" ],
      "matches": [ "*://*.twitch.tv/*" ]
   } ],`

with, ` "content_scripts": [ {
      "js": [ "content.js" ],
      "matches": [ "*://*.twitch.tv/*" ],
      "all_frames":true
   } ],`

I believe that this will allow embedded chats within an iframe to work, solving a bug that has been affecting some users!

tlstommy:
Submitted pull request https://github.com/SevenTV/Extension/pull/919 to add this feature.

Issue #844, opened 2023-08-25
[BUG] not working in iframe twitch chat


Satont:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This issue exists in the latest nightly version

- [X] I am using the latest nightly version

### What browsers are you seeing the problem on?

Firefox

### Current Behavior

7tv not loaded in iframe'd twitch chat.
![image](https://github.com/SevenTV/Extension/assets/42675886/eab32a4e-afdd-46fc-8078-ace3ae5e4122)

Yes, i searched other issues, but seems like they closed as completed.

### Expected Behavior

On this this is loaded ok, so i expect it to load in iframe.
![image](https://github.com/SevenTV/Extension/assets/42675886/d2e1f36b-ea86-4288-834b-d193df8ed2a7)

### Steps To Reproduce

I'm using this code to iframe chat
```vue
<iframe
			v-if="chatUrl"
			:src="`https://www.twitch.tv/embed/${user.login}/chat?parent=${window.location.host}`"
			width="100%"
			height="100%"
			frameborder="0"
			scrolling="no"
			allowfullscreen="true"
		>
		</iframe>
		```

Feridinha:
Any updates on this?

tlstommy:
no.

tlstommy:
I think the issue has to do with "all_frames":true not being in the manifest, but im not sure

tlstommy:
I submitted PR https://github.com/SevenTV/Extension/pull/919 that fixes this bug.

PR #964, opened 2023-12-06
fix: added support for embedded chats

tlstommy:
This is a continuation of a previous [pull request](https://github.com/SevenTV/Extension/pull/919)

I have added `"all_frames":true`, to the manifest for URLs that match a chat embed or popup to support embedded iframes showing 7tv emotes. 

Per Melonifys suggestion I have made it so that it will only work when it detects chat embeds.


tlstommy:
this should fix the following issues: https://github.com/SevenTV/Extension/issues/187 https://github.com/SevenTV/Extension/issues/916 https://github.com/SevenTV/Extension/issues/844

Feridinha:
Dead devs xd

Quoirky:
need that

tlstommy:
Hoping it will be approved @Quoirky! I've got some really useful things I've made for twitch chat that would be even better with 7tv lol!

Feridinha:
> Hoping it will be approved @Quoirky! I've got some really useful things I've made for twitch chat that would be even better with 7tv lol!

Same, I think the extension will not be updated anymore, and they will create a new one from scratch 🤷

lennartkloock:
> > > Hoping it will be approved @Quoirky! I've got some really useful things I've made for twitch chat that would be even better with 7tv lol!
> > 
> > 
> > Same, I think the extension will not be updated anymore, and they will create a new one from scratch 🤷
> 
> Hopefully not, I imagine that they are just undergoing restructuring due to new teams/devs coming in.

I've got to disappoint you. We're currently not planning to continue using or maintaining any of the old code. We're rewriting it completely from scratch in this [monorepo](https://github.com/SevenTV/SevenTV) and fixing literally every bug in the process.
(one of the new devs here btw)

tlstommy:
> > > > Hoping it will be approved @Quoirky! I've got some really useful things I've made for twitch chat that would be even better with 7tv lol!
> 
> > > 
> 
> > > 
> 
> > > Same, I think the extension will not be updated anymore, and they will create a new one from scratch 🤷
> 
> > 
> 
> > Hopefully not, I imagine that they are just undergoing restructuring due to new teams/devs coming in.
> 
> 
> 
> I've got to disappoint you. We're currently not planning to continue using or maintaining any of the old code. We're rewriting it completely from scratch in this [monorepo](https://github.com/SevenTV/SevenTV) and fixing literally every bug in the process.
> 
> (one of the new devs here btw)

Well, in that case I hope you all consider adding this as a feature!

manifest.config.ts
```diff
@@ -50,6 +50,7 @@ export async function getManifest(opt: ManifestOptions): Promise<Manifest.WebExt
 			{
 				matches: ["*://*.twitch.tv/*"],
 				js: ["content.js"],
+				all_frames: true,
```
TroyKomodo, line 53: Does this force a reinstall of the extension?

tlstommy: No, at least as far as I've seen with testing it locally. It just enables 7tv to work in all twitch frames that aren't the top most frame (In this case it's the embedded iframe)

Is that what you're asking?

TroyKomodo: image

No, I am unsure if this constitutes as a change that will uninstall everyone's extensions so I'm scared to make it.

tlstommy: Oh yeah no it won't, at least it shouldn't.

tlstommy: It should just auto update whenever the extension is updated like normal

TroyKomodo: I think, we wont fix this on this version and will wait for the rewrite due to be completed within the next few weeks.

TroyKomodo:
Features against this repo wont be implemented, we are going to rewrite the extension on this [repo](https://github.com/SevenTV/SevenTV).

manifest.config.ts
tlstommy: ok, should I just resubmit a pr then or will you guys plan on adding this feature in?

Issue #973, opened 2023-12-15
[BUG] Suspicious user messages are not highlighted on Twitch


scotsdez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This issue exists in the latest nightly version

- [X] I am using the latest nightly version

### What browsers are you seeing the problem on?

Firefox

### Current Behavior

Even with the "Show Monitored Suspicious User Highlights (Moderator only) " option enabled, messages from users designated as Monitored Suspicious Users are not being highlighted in chat. This behaviour has been observed in 3.0.9 and 3.0.10.1000 on Firefox (currently 120.0.1), however I am aware of other users of the extension who have encountered the same behaviour (unsure of their browser setups).

### Expected Behavior

Messages from Suspicious users should be highlighted in Twitch chat.

### Steps To Reproduce

As a Moderator, access a Twitch channel and open chat. Enable the Show Monitored Suspicious User Highlights (Moderator only) option if not already enabled. Observe chat messages from users designated as monitored suspicious users - they will not be distinguished from non-monitored users.

kabessao:
I'm having the same issue. On 7tv 3.0.16, and firefox 121.0 . This makes the tool difficult to use when moderating, so I'm forced to disable it.

PR #986, opened 2024-01-08
fix(twitch): broken chat embeds

pimothyxd:
If a channel moderator received a "suspicious user" message, the message would not get highlighted. A message containing a clip also seemed to not get embedded. The issue causing this seemed to have been Twitch changing their pubsub client.

See the following examples:

image

which should now look like this:

image

and:

image

should now look like this:

image

Should fix: #973


scotsdez:
@pimothyxd Just in case you hadn't seen the above about the CLA.

Thanks for fixing 973 - I've actually had to stop using 7TV in the meantime as it was making moderation quite difficult, so I'm quite keen to see this drop in a future release.

tlstommy:
> @pimothyxd Just in case you hadn't seen the above about the CLA.
> 
> Thanks for fixing 973 - I've actually had to stop using 7TV in the meantime as it was making moderation quite difficult, so I'm quite keen to see this drop in a future release.

Is the current extension going to be supported while the new version is being built? or will we have to wait for the new version for extension PRs like this one to be reviewed?

TroyKomodo:
Does this still work?
image

pimothyxd:
> Does this still work? image

Should be all good now

PR #990, opened 2024-01-09
fix(twitch): use accent color in chat announcements

pimothyxd:
Fixed a small typo which caused chat announcements to not use the channel's accent color and instead used a white color. See the examples below:

image

which should now look like this:

image

and this is the channel's profile for reference

image


BionicRust:
MahDrilla

Issue #1008, opened 2024-02-29
[BUG] The emote menu button is not seen on the Kick


ozknozsrt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### This issue exists in the latest nightly version

- [X] I am using the latest nightly version

### What browsers are you seeing the problem on?

Chrome

### Current Behavior

The button that opens the emote menu under chat input has disappeared. But tab completion and ":" menu works.

### Expected Behavior

I had to see the emote menu button by the setting button.

![](https://cdn.discordapp.com/attachments/1212499539576229949/1212499539815178320/image.png?ex=65f20f31&is=65df9a31&hm=56cc636cdaa43cf234ff925b4f2feefdd20499bb851fcaf77df442050a0a4f17&)

### Steps To Reproduce

Maybe the Kick has changed the relevant html tag.

![](https://images-ext-2.discordapp.net/external/0h8Dt9aluN2I8vSZS-q2M0_7GlJ-loN9JLa6Za8RSEI/https/img001.prntscr.com/file/img001/ayqZFH-6R2CVZE6AHOzR-g.png?format=webp&quality=lossless)

I guess it can't find this div. (file path: [Extension/src/site/kick.com/modules/chat/ChatEmoteMenu.vue](https://github.com/SevenTV/Extension/blob/a98e3653a0d0714f945e57d20daf53f2cf5834f0/src/site/kick.com/modules/chat/ChatEmoteMenu.vue#L45))

![](https://media.discordapp.net/attachments/1212499539576229949/1212503505236140042/image.png?ex=65f212e2&is=65df9de2&hm=a579f2d4de9fafb36cc1c34829258611d5c386a0b1a819ca8d37242816e72142&=&format=webp&quality=lossless)

ozknozsrt:
The problem solved. I mark it as completed

PR #1010, opened 2024-03-01
fix(kick): emote menu button

pimothyxd:
## Proposed changes

A simple UI update by Kick changed the chat input container where the emote picker icon was being placed. This PR simply updates it to the new one. 

Additionally, the previous implementation never reset the channel state after leaving a channel. So re-entering the same channel (e.g. by going to the homepage and back) would cause the emote picker to also not rerender. Along with that change, if no message was dispatched by the room it would also cause the channel to never be assigned. These are both fixed by this PR.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Further comments

A snippet of what it was before:

logged_in

and after:

logged_in

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -17,6 +17,7 @@
 -   Added an option to hide timestamps in vods
 -   Fixed an issue that caused replies in threads to not appear
 -   Fixed an issue where replies in threads could not be selected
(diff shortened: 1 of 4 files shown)


ozknozsrt:
related issue #1008

PR #1051, opened 2024-06-20
Feat: Animated FFZ emote support

ChrisScr3ams:
## Proposed changes

Only static FFZ emotes are currently supported by 7TV. For animated emotes only the static version is used, probably because FFZ itself only added support for them rather recently. Because of that 7TV should add support for animated FFZ emotes as well.

## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Further comments

The FFZ API has a field "animated" which returns a map of DPI scales to animated emote URLs and is null if the emote is not animated. This field was added to our interface and the transformer adjusted to use the correct url and file dependant on if the emote is animated or not.

Documentation from FFZ API: 
Screenshot 2024-06-20 203722

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -1,6 +1,9 @@
-### 3.1.1.2100
+### 3.2.2 1000
 
 -   Added support for animated FFZ emotes
+
+### 3.1.1.2100
+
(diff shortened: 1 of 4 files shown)


Excellify:
Mind switching changelog version to 3.1.1.3000 and dev version to 3.0?

PR #1053, opened 2024-07-04
Feat: added support for pop-out and embedded chats.

tlstommy:
## Proposed changes
When viewing a Twitch chat that is using a Twitch chat embed url (`https://www.twitch.tv/embed/<channel>/chat?`) or a popout chat url (`https://www.twitch.tv/popout/<channel>/chat`), the 7TV extension will not load or work because it is only allowed to access top-level windows. 

This is a simple fix that enables `"all_frames": true` in the manifest and then adds checks to see if the current window is a regular top-level window or a chat pop-out or embed. 

This solves the following issues: https://github.com/SevenTV/Extension/issues/916 https://github.com/SevenTV/Extension/issues/844 https://github.com/SevenTV/Extension/issues/187

## Types of changes

## Checklist

- [x] Lint and unit tests pass locally with my changes

## Further comments

If this is a relatively large or complex change, you may want to start the discussion by explaining why you chose the solution you did, what alternatives you considered, etc.

### DIFF
--- a/CHANGELOG-nightly.md
+++ b/CHANGELOG-nightly.md
@@ -3,6 +3,7 @@
 -   Added support for animated FFZ emotes
 -   Added option to hide whispers
 -   Added tooltip to error messages in /search
+-   Added support for 7TV to work in Twitch chat embeds and popouts
 
 ### 3.1.1.2100
 
--- a/manifest.config.ts
+++ b/manifest.config.ts
@@ -50,6 +50,7 @@ export async function getManifest(opt: ManifestOptions): Promise<Manifest.WebExt
 			{
 				matches: ["*://*.twitch.tv/*"],
 				js: ["content.js"],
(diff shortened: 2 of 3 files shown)


TroyKomodo:
I suspect this will require a store release @Excellify, if so I dont think we can merge this. 

For OP; we have 2 deployment methods on the extension. 1. A hosted release, a store release.  We cannot make store releases without first fixing some other issues with the chrome store / firefox store. 

We plan to redo the extension entirely and we do not have the bandwidth to fix the extension's store issues. However this might be able to deployed as a tapermonkey script.

Excellify:
We can't add permissions to the manifest either as that would cause the extension to be disabled for everyone who has it installed. We can only add optional permissions and request those at runtime.

tlstommy:
I understand, thanks!

PR #1121, opened 2025-01-08
Added an 'always on' emote autocompletion option

ilianoKokoro:
## Proposed changes

This changes the colon autocompletion to add an 'always on' option. When enabled, the emote autocompletion menu opens even when the ':' prefix is not typed. This is useful if the user types a lot of emotes and wants very fast access to them.
Also, this is a feature in FFZ for example.

## Types of changes

## Checklist

(I could not sign the CLA)
image
- [x] I have added necessary documentation (if appropriate)
(N/A)
- [x] Any dependent changes have been merged
(N/A)


src/site/twitch.tv/modules/chat-input/ChatInput.vue
```diff
@@ -161,6 +174,10 @@ function findMatchingTokens(str: string, mode: "tab" | "colon" = "tab", limit?:
 	return matches;
 }
 
+onMounted(() => {
+	window.addEventListener("keydown", handleCapturedKeyDown, { capture: true });
```
Excellify, line 178: Can use vueuse's useEventListener here to automatcally handle mount and unmount

src/site/twitch.tv/modules/chat-input/ChatInput.vue
```diff
@@ -406,14 +423,78 @@ function onKeyDown(ev: KeyboardEvent) {
 	}
 }
 
+function handleCapturedKeyDown(ev: KeyboardEvent) {
+	// Prevents autocompletion on Enter when completion mode is -> always on
+	if (ev.key === "Enter") {
+		const { component } = props.instance;
+		const { componentRef } = component;
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		const componentProps: any = componentRef.props;
```
Excellify, line 432: Please add typings for whatever props youre using

src/site/twitch.tv/modules/chat-input/ChatInputModule.vue
```diff
@@ -130,24 +130,29 @@ markAsReady();
 
 <script lang="ts">
 export const config = [
-	declareConfig("chat_input.autocomplete.colon", "TOGGLE", {
```
Excellify, line 133: Please dont rename config keys as that would cause users to lose their current configuration.
If youre changing the scheema please add a transform converting the old value to fit the new options.

Excellify (changes requested):

src/site/twitch.tv/modules/chat-input/ChatInputModule.vue
ilianoKokoro: That makes sense. I changed it because the name of the config would make less sense with the always on option. I'll change it back

src/site/twitch.tv/modules/chat-input/ChatInput.vue
ilianoKokoro: It doesn't seem to work with `useEventListener` because it has no capture property. The enter event reaches before it can be intercepted. I can implement onUnmounted manually.

ilianoKokoro: Nevermind I made it work with `window`. I had `document` targeted.

src/site/twitch.tv/modules/chat-input/ChatInput.vue
ilianoKokoro: image

ilianoKokoro: Alright I think I found the right way to type it. lmk if that's right

